PR #57, opened 2015-04-17
Add speed in bps for pkt-gen because it's useful to see traffic speed

pavel-odintsov:
Hello!

Please review my patch.


vmaffione:
The funcionalities referred here have already been merged, so this is obsolete and out of date.

PR #87, opened 2015-09-23
NR_REG_SINGLE_RX: single RX queue mode support

jibi:
In current modes (`NR_REG_ONE_NIC`, `NR_REG_ALL_NIC`) Netmap takes over all RX queues.
For some applications this is undesirable: the user might want to enable Netmap only on a single RX queue and leave the rest untouched (i.e. all packets flowing to other queues are handled as usual: pushed to the kernel without any Netmap interference).

This patch introduces support for a new mode called single RX queue (`NR_REG_SINGLE_RX`), in which an adapter exposes only the requested RX rings to the Netmap applications.
All the other RX queues are kept attached to the host stack, thus keeping all the network flows working as usual. The TX side is as well managed by the kernel, untouched by Netmap.

This mode can be useful for applications receiving packets at high speed, when it's not affordable to put an entire NIC in Netmap mode. We believe this approach avoids the cost and complexity of the transparent mode.

An example use would look like:

* With an indirection table manipulation user isolates one RX queue from normal network traffic.
* User adds a dedicated flow steering rule to direct only a specific network flow to the isolated RX queue.
* User opens an interface in `NR_REG_SINGLE_RX` mode on the isolated queue and is able to receive packets at high speed using a well defined Netmap APIs.
* All other network flows going to the kernel are uninterrupted and do not need any support from the Netmap application.

For ease of use we propose the following syntax support for the single rx queue mode in the netmap_user.h helpers (X is the RX queue number):

```c
nm_open("netmap:ifN~X", NULL, 0, 0)
````

It's worth noting three quirks:
* The transmission side is not handled by this patch. It is not possible to access the TX ring from the application that has opened the interface in `NR_REG_SINGLE_RX` mode.
* While in `NR_REG_SINGLE_RX` mode, the adapter cannot be reopened in a different mode, and viceversa, while already opened in a different mode, the adaptor cannot be reopened in `NR_REG_SINGLE_RX` mode. 
* Some hardware RX queues are special. For example ARP packets are often pushed to RX queue #0. Applications using single RX mode with RX queue #0 might need to handle ARP protocol.

### DIFF
--- a/sys/dev/netmap/netmap.c
+++ b/sys/dev/netmap/netmap.c
@@ -3094,7 +3094,7 @@ netmap_reset(struct netmap_adapter *na, enum txrx tx, u_int n,
  * - for a nic connected to a switch, call the proper forwarding routine
  *   (see netmap_bwrap_intr_notify)
  */
-void
+int
 netmap_common_irq(struct netmap_adapter *na, u_int q, u_int *work_done)
 {
 	struct netmap_kring *kring;
@@ -3107,7 +3107,7 @@ netmap_common_irq(struct netmap_adapter *na, u_int q, u_int *work_done)
 	}
 
 	if (q >= nma_get_nrings(na, t))
-		return;	// not a physical queue
+		return 0; // not a physical queue
 
 	kring = NMR(na, t) + q;
 
@@ -3116,6 +3116,7 @@ netmap_common_irq(struct netmap_adapter *na, u_int q, u_int *work_done)
 		*work_done = 1; /* do not fire napi again */
 	}
 	kring->nm_notify(kring, 0);
+	return 1;
 }
 
 
@@ -3154,8 +3155,7 @@ netmap_rx_irq(struct ifnet *ifp, u_int q, u_int *work_done)
 		return 0;
 	}
 
-	netmap_common_irq(na, q, work_done);
-	return 1;
+	return netmap_common_irq(na, q, work_done);
 }
 
 
--- a/sys/dev/netmap/netmap.c
+++ b/sys/dev/netmap/netmap.c
@@ -1730,6 +1730,12 @@ netmap_interp_ringid(struct netmap_priv_d *priv, uint16_t ringid, uint32_t flags
 	case NR_REG_PIPE_MASTER:
 	case NR_REG_PIPE_SLAVE:
 		for_rx_tx(t) {
+			if ((t == NR_RX && (flags & NR_TX_RINGS_ONLY)) ||
+			    (t == NR_TX && (flags & NR_RX_RINGS_ONLY))) {
+				priv->np_qfirst[t] = priv->np_qlast[t] = 0;
+				continue;
+			}
+
 			priv->np_qfirst[t] = 0;
 			priv->np_qlast[t] = nma_get_nrings(na, t);
 		}
@@ -1743,6 +1749,18 @@ netmap_interp_ringid(struct netmap_priv_d *priv, uint16_t ringid, uint32_t flags
 			return EINVAL;
 		}
 		for_rx_tx(t) {
+			if ((t == NR_RX && (flags & NR_TX_RINGS_ONLY)) ||
+			    (t == NR_TX && (flags & NR_RX_RINGS_ONLY))) {
+				/*
+				 * NR_[TX|RX]_RINGS_ONLY affects only hw
+				 * rings, so keep sw rings on in any
+				 * case
+				 */
+				priv->np_qfirst[t] = nma_get_nrings(na, t);
+				priv->np_qlast[t] = nma_get_nrings(na, t) + 1;
+				continue;
+			}
+
 			priv->np_qfirst[t] = (reg == NR_REG_SW ?
 				nma_get_nrings(na, t) : 0);
 			priv->np_qlast[t] = nma_get_nrings(na, t) + 1;
@@ -1756,6 +1774,12 @@ netmap_interp_ringid(struct netmap_priv_d *priv, uint16_t ringid, uint32_t flags
 			return EINVAL;
 		}
 		for_rx_tx(t) {
+			if ((t == NR_RX && (flags & NR_TX_RINGS_ONLY)) ||
+			    (t == NR_TX && (flags & NR_RX_RINGS_ONLY))) {
+				priv->np_qfirst[t] = priv->np_qlast[t] = 0;
+				continue;
+			}
+
 			/* if not enough rings, use the first one */
 			j = i;
 			if (j >= nma_get_nrings(na, t))
--- a/sys/dev/netmap/netmap.c
+++ b/sys/dev/netmap/netmap.c
@@ -815,6 +815,8 @@ netmap_krings_create(struct netmap_adapter *na, u_int tailroom)
 			kring->ring_id = i;
 			kring->tx = t;
 			kring->nkr_num_slots = ndesc;
+			kring->nr_mode = NKR_NETMAP_OFF;
+			kring->nr_pending_mode = NKR_NETMAP_OFF;
 			if (i < nma_get_nrings(na, t)) {
 				kring->nm_sync = (t == NR_TX ? na->nm_txsync : na->nm_rxsync);
 			} else {
@@ -909,11 +911,20 @@ static void
 netmap_do_unregif(struct netmap_priv_d *priv)
 {
 	struct netmap_adapter *na = priv->np_na;
+	enum txrx t;
+	int i;
 
 	NMG_LOCK_ASSERT();
 	na->active_fds--;
 	/* release exclusive use if it was requested on regif */
 	netmap_rel_exclusive(priv);
+
+	for_rx_tx(t) {
+		for (i = priv->np_qfirst[t]; i < priv->np_qlast[t]; i++) {
+			kring_rel_netmap_mode(priv, t, i);
+		}
+	}
+
 	if (na->active_fds <= 0) {	/* last instance */
 
 		if (netmap_verbose)
@@ -948,6 +959,10 @@ netmap_do_unregif(struct netmap_priv_d *priv)
 		/* delete rings and buffers */
 		netmap_mem_rings_delete(na);
 		na->nm_krings_delete(na);
+	} else {
+		if (kring_pending(priv)) {
+			na->nm_register(na, 1);
+		}
 	}
 	/* possibily decrement counter of tx_si/rx_si users */
 	netmap_unset_ringid(priv);
@@ -1971,6 +1986,8 @@ netmap_do_regif(struct netmap_priv_d *priv, struct netmap_adapter *na,
 {
 	struct netmap_if *nifp = NULL;
 	int error;
+	enum txrx t;
+	int i;
 
 	NMG_LOCK_ASSERT();
 	/* ring configuration may have changed, fetch from the card */
@@ -2011,6 +2028,12 @@ netmap_do_regif(struct netmap_priv_d *priv, struct netmap_adapter *na,
 	if (error)
 		goto err_del_rings;
 
+	for_rx_tx(t) {
+		for (i = priv->np_qfirst[t]; i < priv->np_qlast[t]; i++) {
+			kring_get_netmap_mode(priv, t, i);
+		}
+	}
+
 	/* in all cases, create a new netmap if */
 	nifp = netmap_mem_if_new(na);
 	if (nifp == NULL) {
@@ -2032,6 +2055,10 @@ netmap_do_regif(struct netmap_priv_d *priv, struct netmap_adapter *na,
 		error = na->nm_register(na, 1); /* mode on */
 		if (error) 
 			goto err_del_if;
+	} else {
+		if (kring_pending(priv)) {
+			na->nm_register(na, 1);
+		}
 	}
 
 	/*
@@ -2050,6 +2077,12 @@ err_del_if:
 	netmap_mem_if_delete(na, nifp);
 err_rel_excl:
 	netmap_rel_exclusive(priv);
+
+	for_rx_tx(t) {
+		for (i = priv->np_qfirst[t]; i < priv->np_qlast[t]; i++) {
+			kring_rel_netmap_mode(priv, t, i);
+		}
+	}
 err_del_rings:
 	if (na->active_fds == 0)
 		netmap_mem_rings_delete(na);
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by pavel-odintsov** [comment]
Great job! With this patch we could run netmap-ipfw on any Linux box and use it for traffic filtration on line rate / 10 GE.

File: sys/dev/netmap/netmap.c [hunk 1]
```diff
@@ -3079,15 +3150,26 @@ netmap_common_irq(struct ifnet *ifp, u_int q, u_int *work_done)
 	}
 
 	if (q >= nma_get_nrings(na, t))
-		return;	// not a physical queue
+		return 0; // not a physical queue
```
john-gress: If you are using the netmap patched ixgbe drivers, this will cause the driver to fall back to regular processing.

pavel-odintsov:
Any news? Really awesome code and features. What about upstream merge?

giuseppelettieri:
Sorry guys, just starting to look at the code. I personally like the idea, but I would go for something more orthogonal. I would have a NETMAP_ON flag on each ring (preferably on the kring and not on a separate bitmap). When you open netmap:ethX-NN you only put the queue pair NN in netmap mode and all other queues do not change their mode. This is a change of API, so maybe we need a new flag. If, in addition, you only want to open the RX queue in the pair, you add another flag after '/'.

Internally, I would refactor the code so that everything is per-ring. netmap_transmit should always be there and fall back to the saved nic callback if the ring is not in netmap mode. netmap_do_unregif should decrement the per-ring reference counter and call ->nm_register() if any ring must return to normal mode, and so on. In general, I would like to avoid the nm_single_rx_mode flag and all the special cases that it introduces in the code.

Do you see any problem with any of this?

jibi:
Sounds sane. So, to sum up:

* remove the bitmap and use a field inside `netmap_kring` to determine if the ring is or not in nm mode (so we do reference count on each ring)
* do not overwrite the `netdev_ops` struct but decide inside `netmap_transmit` if we should pass the control to the OS
* use a more generic API: basically reuse the `iface-queue` syntax, adding an additional flag after `/` to request only the RX/TX ring
* if one of the ring changes its state from NETMAP_ON/NETMAP_OFF in `netmap_do_regif` or `netmap_do_unregif`, call `netmap_reset` on the adaptor

Do I get it right?

giuseppelettieri:
Yes, exactly. Only, I am not sure on your second point: we need to overwrite the netmap_ops if we want netmap_transmit to get called, don't we? What I am saying is that we should overwrite netmap_ops regardless of the netmap mode, when the first ring changes its state from NETMAP_OFF to NETMAP_ON. Then, it is as you say: netmap_transmit decides whether it should pass the control to the OS, or not, based on the per-ring flags.

jibi:
Yes sure, what I meant was: do not overwrite `netmap_ops` with the OS callback when we are in single queue mode, always use `netmap_receive`, and let it decide what to do.

That said, thanks for the advices.
I will start working on, so expect a PR update soon.

jibi:
I just pushed the current WIP patch (just want to make sure I'm going into the right direction).
Here are some notes.

New flags introduced: `NR_TX_RINGS_ONLY` and `NR_RX_RINGS_ONLY`, which are used to access only the TX or RX rings (i.e. the rings that are not requested will remain in normal mode).

New fields in the `netmap_kring` struct:
* `nr_mode`: the current mode of the ring (can be either `NETMAP_ON` or `NETMAP_OFF`)
* `nr_pending_mode`: the pending mode of the ring. After `netmap_reset` is called, this will become the effective ring mode
* `nr_refcount`: reference counter for the ring's users

#### How does it work
`netmap_interp_ringid` has been modified to take into account the `NR_TX_RINGS_ONLY` and `NR_RX_RINGS_ONLY` flags. If one of these flags is specified, the `np_qfirst` and `np_qlast` variables for the specified ring tipology are set to 0.

After this, for each ring that needs to be put in netmap mode (information that we obtain from the `np_qfirst` and `np_qlast` variables), `kring_get_netmap_mode` is called from `netmap_do_regif`:

```c
        for_rx_tx(t) {
                for (i = priv->np_qfirst[t]; i < priv->np_qlast[t]; i++) {
                        kring_get_netmap_mode(na, i, t);
                }
        }
```
this function just sets `nr_pending_mode` to `NETMAP_ON`, and increments the reference counter.
If the ring is physical, `nr_mode` will be set to `NETMAP_ON` only after `netmap_reset` has been called. Otherwise (i.e sw ring), the mode is set directly in this function.

After this, if there are any pending rings (i.e. `nr_mode` != `nr_pending_mode`), `nm_register` (and in turn `netmap_reset`) are called.

When an interface is closed, `netmap_do_unregif` is called, which in turn will call `kring_put_netmap_mode` on each ring that belongs to that priv interface. This function just decrements the reference counter for the ring, and if this reaches 0, it sets the `nr_pending_mode` to `NETMAP_OFF` (again, `nr_mode` will be set to `NETMAP_OFF` only after `netmap_reset` has been called).

`netmap_common_irq` is where we check the ring mode, returning either 0 (to signal that netmap is not handling this particular ring so that the NIC driver can do it) or 1 (which is, netmap handled the IRQ).

Regarding the TX bypass (i.e. allowing the OS to use physical TX rings to transmit packets), I used the sw ring as a switch:
* if the user does not request the TX software/host ring (either by requesting only the hw rings or by specifying `NR_RX_RINGS_ONLY`), `netmap_transmit` will call the original OS callback to deliver packets to the NIC (for example the `ndo_start_xmit` in the Linux case).
* otherwise everything will work as usual (i.e. when a packet need to be transmitted, `netmap_transmit` is called and the packet is written into the netmap host ring).

#### Isssues
The only flaw of this design is that at the moment, an user could request a mode which would not make sense: if we open an interface in hardware mode (i.e. we request access to all the physical rings), the host ones will be not handled by netmap, which means that when the network stack needs to send a packet, it will just call the OS xmit callback. But in this case there would be a contention on the physical TX rings, since both the userspace application and the OS are writing packets on those rings.

giuseppelettieri:
The current direction looks good. Here is a few comments:
- first, the issue you mention at the bottom. I think that, whatever the open mode, we should always end up with each NIC ring assigned to either netmap or the host stack.  The NETMAP_ON state of the host rings should not depend on whether a user has actually binded them or not, but only on the NETMAP_ON state of the other rings. To make things simple, we may NETMAP_ON  them in all modes. 
- there already is a 'users' field in the krings, used to implement the exclusive flag (see netmap_{get,rel}_exclusive). Do you think we may reuse that for the reference count?

jibi:
> we should always end up with each NIC ring assigned to either netmap or the host stack

I agree, and that's how it works now: a ring is either in `NETMAP_ON` or `NETMAP_OFF` mode.

For **hw** rings (both RX and TX) this works ok:
* if a ring is in nm mode, all IRQs for this ring are handled by netmap
* otherwise the NIC driver has to take care of the IRQs

For **RX sw** ring this works as well:
* if this ring is in nm mode, a userspace program can write it and send packets to the host stack
* otherwise a userspace program simply cannot access that ring (so, no bad things happen)

For **TX sw** ring:
* if this ring is in nm mode, when the kernel has some packets to send, it will write them to this ring, and a userspace program can access these packets
* otherwise we have a sort of "bypass TX" semantic, i.e. all packets sent by the kernel are actually written on the hw TX rings and sent (using an OS function). Again, for this mode to work, hw TX rings must be in normal mode otherwise, as far as I know (and that's my main concern), a userspace program could access an hw TX ring which already contains packets written by the kernel.

> The NETMAP_ON state of the host rings should not depend on whether a user has actually binded them or not, but only on the NETMAP_ON state of the other rings. 

I'm not following you. Do you mean, put host rings in netmap mode only if all the other (hw) rings are in netmap mode?

> To make things simple, we may NETMAP_ON them in all modes

If we `NETMAP_ON` them in all modes, there's no way to specify the "bypass TX" semantic (and that's one of the main points of the "single-rx-queue" mode: the ability to access one RX queue while keeping everything else working in normal mode)

> there already is a 'users' field in the krings, used to implement the exclusive flag (see netmap_{get,rel}_exclusive). Do you think we may reuse that for the reference count?

Sure, I had a look at the code and it seems fairly simple to use that as reference counter.

giuseppelettieri:
OK, I think I see where the confusion lies. I was talking about the NIC rings, not the netmap rings. Each *NIC* ring should be either "assigned to netmap" or "assigned to the host stack":
- *assigned to netmap*: there is a netmap hw ring that netmap applications use to exchange frames with the NIC ring;  for tx rings, packets coming from the host stack and headed towards this NIC ring must go to the netmap sw rx (note the switch from tx to rx) ring (aka host rx ring) instead, regardless of whether some application has opened the host rx ring or not; for rx rings, only netmap sees the frames.
- *assigned to the host (stack)*: the NIC ring is invisible to netmap applications, there is no netmap hw ring that exchanges frames with this NIC ring; for tx, frames coming from the host stack and headed towards this NIC ring actually go to this NIC ring; for rx, frames received on this ring follow their normal path to the host stack.

Netmap hw rings have a 1-1 mapping with NIC rings, but netmap sw (aka host) rings have not. 
- The netmap sw rx ring, as of now, collects all the frames that the host stack wants to send through the NIC when this is put in netmap mode. In the new scheme, it should collect only the frames that the host stack wants to send through NIC rings that have been put in netmap mode. 
- The netmap sw tx ring is used by netmap applications to send packets to the host stack and let the host stack think that the frames actually come from the NIC. Maybe we can let this functionality as always available.

Maybe we should look at all possibilities and see if they make sense. This is how it works now:
[1] netmap:iface -> all NIC rings are put in netmap mode; app has registered the netmap hw rings only;
[2] netmap:iface^ -> all NIC rings are put in netmap mode; app has only registered the sw rings;
[3] netmap:iface* -> all NIC rings are put in netmap mode; app has registered all netmap rings (both hw and sw)
[4] netmap:iface-NN -> all NIC rings are put in netmap mode; app has only registered for one hw ring pair (tx and rx)

In the new scheme, [4] should put NETMAP_ON only on hw ring pair NN (both tx and rx). Note that this means that the netmap sw rings are now ON too, even if this app has not opened them (it may open them with another file descriptor, though). [1] and [3] should work as before, but I am not sure about what we should do in [2]. Suggestions welcome. Moreover, even in [4], this new behaviour should be probably only triggered by a flag. Again, not sure: maybe it actually makes more sense in the new scheme.

Note that all of this is independent from the new R and T flags, which add yet another dimension. With /R we put only the rx NIC ring(s) in netmap mode, and only the NIC tx ring(s) with /T. Note again: putting any NIC tx ring in netmap mode automatically enables the netmap sw rx ring, since we need a place to put the host stack frames that would have otherwise landed on that ring, even if no app has registered the host ring.

jibi:
Good, now I get it.
Basically, if a packet (from the host stack) is going to end up in a TX ring which is in Netmap mode, we should move that packet into the RX host ring.
After that we shoud figure out the various edge cases.
One thing that is sure is: having at least one hw TX ring in Netmap mode requires to have the RX sw ring as a "dependency" (a place where we copy packets that would go to that ring).

Regarding the hw TX to sw RX change, in Linux this should not be too difficult to implement.
As far as I know, the `sk_buff` struct holding the packet that is going to be transmitted contains a field (`queue_mapping`) which indicates the TX queue number that has to be used for that packet.
So it would be just a matter of checking inside `netmap_transmit` if `queue_mapping` is in netmap mode (in that case we should just continue executing `netmap_transmit`) or not (in that case we would call the original `ndo_start_xmit` function).

I don't know if this is the case also with FreeBSD.

jibi:
Here's the updated PR, and some comments about the changes.

#### struct netmap_kring
This struct has gained two fields:

```c
	uint32_t	nr_mode;
	uint32_t	nr_pending_mode;
```

`nr_mode` is the actual ring mode, `nr_pending_mode` is the mode that will take effect after `netmap_reset` is called on that ring.

#### rings mode setup
The `np_qfirst` and `np_qlast` fields in the `netmap_priv_d` struct are still used to keep track of the requested rings, but they are used only for the **hw** ones.
There is another variable (`np_sw_rings`) that is used to tell if sw rings are actually on or off (they are turned off only for the PIPE_\* modes).

After `netmap_interp_ringid` has been called, according to `np_qfirst`, `np_qlast` and `np_sw_rings`, `kring_get_netmap_mode` is called on every ring that needs to be in netmap mode. 

This function just sets the **pending** mode to netmap on, so that `netmap_reset` can later actually enable the ring.

After that, `netmap_get_exclusive` is called, which will increment the krings `users` counter.

#### iterating rings
Since `np_qfirst` and `np_qlast` do not take into account the sw rings, the usual loop

```c
for (i = priv->np_qfirst[NR_TX]; i < priv->np_qlast[NR_TX]; i++)
```

has been replaced with the macro `for_each_priv_ring`, which takes into account also sw rings:

#### transmitting with hw rings

`netmap_transmit` has been modified to check if a packet can be transmitted using a hw TX ring:

* if the packet is going to use a TX ring which is not in netmap mode, `netmap_transmit` will call the OS callback to transmit it using the a hw ring
* otherwise `netmap_transmit` will continue executing and the packet will be enqueued in the RX host ring (this was tested only on Linux)

#### si_users
The optimization that wakes up only one of the users (if this is listening only on one queue) has been removed, since now there are always at least two queues, even in single ring pair mode (one hw and one sw).

#### testing
The patch has been tested with two example programs.

With the first one I tested the transparent mode: https://gist.github.com/jibi/fce7f3258ae1af5938fc

This program forwards (setting the slot flag to `NS_FORWARD`) each packet read from an RX ring (hw or sw).

The second program was used to test the single ring pair mode: https://gist.github.com/jibi/abef9306f6ccb1628617

Multiple instances of the test have been launched on different rings while the machine was receiving a large number of HTTP requests (so I actually verified that the packets that were going to be transmitted by a TX ring in netmap mode were copied to the RX sw ring).

pavel-odintsov:
Great! Awesome! ;)

nyrahul:
Have a doubt with regards to handling of IP fragmentation with this PR.

Usually an RSS queue with L4 tuple will result in all the IP-fragmented packets to go to RX queue  0 and non-IPfragmented packets to go to some other queue[1].
Wonder if this happens and if some of the packets in the same TCP stream are mapped to different queues, then it would be total stream failure. Wont it?

In general with RSS queues and assuming all the queues handled by kernel, this may result in out-of-order processing of certain packets and may lead to sub-optimal performance (but atleast it will work). But with the use of netmap + linux kernel in parallel, the streams would totally fail. Isn't it?

Ref:
[1] http://adrianchadd.blogspot.in/2014/08/receive-side-scaling-figuring-out-how.html

vmaffione:
You are right, indeed RSS needs to be carefully programmed taking into account these problems.
A wrong RSS configuration with netmap reading from a subset of the queues can easily lead to a broken system.
You can choose whether to use this feature or not (e.g. opening all the RX queues in netmap mode) depending on your needs.

xfliu2665:
I wonder if there is a way could realize in netmayp, which i register a filter callback in netmap, then netmap use the filter to determine to pass the frame to userspace,or pass it to kernel.  i mean when the filter return a true, netmap pass it to userspace, pass to kernel otherwise.

i really look forward to netmap could support it!

PR #91, opened 2015-10-06
linux/virtio: add missing sg_init_table()

kalyazin:
rx and tx scatterlists should be initialized before passing them to
virtqueue helpers, because the helpers might detect list chaining which
is not expected (virtio netmap uses one buffer per packet).

Signed-off-by: Nikita Kalyazin <n.kalyazin@samsung.com>

### DIFF
--- a/LINUX/virtio_netmap.h
+++ b/LINUX/virtio_netmap.h
@@ -114,6 +114,8 @@ static void free_receive_bufs(struct SOFTC_T *vi)
 #define GET_RX_SG(_vi, _i)		(_vi)->rq[_i].sg
 #define GET_TX_SG(_vi, _i)		(_vi)->sq[_i].sg
 #define COMPAT_DECL_SG
+#define INIT_SGS(_vi)			virtio_netmap_init_sgs(_vi)
+#define SG_INIT_TABLE(_sgl, _n)
 #ifdef NETMAP_LINUX_VIRTIO_RQ_NUM
 /* multi queue, num field exists */
 #define DECR_NUM(_vi, _i)		--(_vi)->rq[_i].num
@@ -141,6 +143,8 @@ static void free_receive_bufs(struct SOFTC_T *vi)
    function. This macro does this definition, which is not necessary
    for subsequent versions. */
 #define COMPAT_DECL_SG			struct scatterlist _compat_sg;
+#define INIT_SGS(_vi)
+#define SG_INIT_TABLE(_sgl, _n)		sg_init_table(_sgl, _n)
 /* Use the scatterlist struct defined in the current function */
 #define GET_RX_SG(_vi, _i)	&_compat_sg
 #define GET_TX_SG(_vi, _i)	&_compat_sg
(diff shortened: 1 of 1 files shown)


vmaffione:
RIght, but why did you come up with this specific issue? Did you experience any problems?
In any case, it's not necessary to do the initialization of the sg array each time sg_set_buf() is called. You can call sg_init_table() once for all the queues, at netmap register time.

kalyazin:
Exactly, I observed adding two buffers instead of one. I presume, it happens because host driver usually uses two buffers per packet, and sg structure remains same.
I experienced this when trying to use the following OpenDataPlane patches for netmap-virtio:
https://lists.linaro.org/pipermail/lng-odp/2015-October/015753.html
Indeed, I was thinking about calling sg_init_table() only once, but wasn't sure I would do it right. Anyway, I'm going to implement this improvement.

kalyazin:
I've uploaded new version of the patch. Does it look better?

vmaffione:
Yes, thanks, but unfortunately this will break in this case
 
https://github.com/luigirizzo/netmap/blob/e60212c3db1a90543646dce7929b60d95ceeb078/LINUX/virtio_netmap.h#L143

e.g. for linux < 2.6.35.
So in this very case  (and only this case) we do need sg_init_table for each packet, since the sg is a local variable. It is necessary to play with macros here. We need to check the compilation for all the linux versions maintained in netmap (from 2.6.32 on).

kalyazin:
Thanks for the comment.
Pushed updated version.

vmaffione:
Ok, thanks. Testing compilation, it will take a while.
With this patch, are you noticing difference in performance w.r.t the current version?
In the current version, each frame should take up 19 virtio descriptors, while with this patch each frame should take only 2. So it could be the case that the performance increases.

kalyazin:
I'm not sure I follow you. The version in master does not work for me at all, since it gives wrong information about the packet it wants to send (my back-end thinks the packet contains more than one segment, which is not the case).
What do you mean by 19 descriptors vs 2? Sorry, I don't get where number 19 comes from.

When it comes to comparison, I did some measurements regarding moving from 2 descriptors per packet to 1 descriptor. I'm not sure my measurement is relevant enough, because it doesn't use netmap at all. I have virtio-user front-end and vhost-user back-end. With 1 descriptor per packet I have about 15% performance boost over 2 descriptors.

Btw, do you have some continuous integration system to check the build status for all supported kernel versions or build it by hand?

vmaffione:
The usual tests I do (for TX side) is to let a QEMU VM transmit using pkt-gen netmap example, over a virtio-net adapter.

This said, in a completely different context (not netmap), where virtio-net was involved, I figured out that virtio-net can use sg lists whose length is up to 19, where 19 comes from MAX_SKB_FRAGS+2

http://lxr.free-electrons.com/source/drivers/net/virtio_net.c#L71

In my experiment, I was seing the virtio-net backend consuming 19 descriptors for each processed packet, instead of 1 or 2 (as I thought). After some days, this issue was opened, and I suddenly realized that the same could happen with netmap - we are using 19 per-packet rather than 2. I still have to verify it, however.

Yes, we have an helper for integration testing, it's in LINUX/scripts/np

kalyazin:
Ah, I now see about 19.
I suppose, the formula "65536/PAGE_SIZE + 1" has been invented for TSO packets whose maximum size is 64K. So one has good chances to observe sg list of length 19, only if she enables TSO feature.
However, as I understand, netmap (even current implementation) allows to send non-segmented packets only, because it always sets one buffer and adds it to virtqueue for each netmap ring slot:
```
sg_set_buf(sg, addr, len);
err = virtqueue_add_outbuf(vq, sg, 1, na, GFP_ATOMIC);
```
So I don't see how netmap could leverage more than 1 descriptor (I mean, in correct operation). That's why I didn't expect any effect on performance from the patch in question.

vmaffione:
Hi @kalyazin , I've merged your commit in our internal repository, together with other rearrangements and fixes to achieve compilation on all the supported versions. The change will be available on this public repo soon. Sorry that this took so much time, but I didn't have any chance to work on it. Thanks for your contribution.

kalyazin:
Hi @vmaffione ,
Thanks.

PR #92, opened 2015-10-06
linux/virtio: configure real number of rx queues

kalyazin:
Signed-off-by: Nikita Kalyazin <n.kalyazin@samsung.com>


rbtcollins:
This patch seems interesting, but the delta seems to include wholesale deletion of a large chunk of the tree - perhaps it needs rebasing? [Not a committer, just looking around for patches that may help my bug :)]

vmaffione:
Yes, this needs rebasing. The patch itself is straightforward

https://github.com/luigirizzo/netmap/pull/92/commits/f35330d5be0c6286499197c13e7c401cfaa002b8

but it needs to be tested and I need to be able to reproduce a scenario in which it is working.

kalyazin:
I rebased the patch.

I used OpenDataPlane netmap driver for testing when developed the patch:
https://git.linaro.org/lng/odp.git/blob/HEAD:/platform/linux-generic/pktio/netmap.c
"switch" example (https://git.linaro.org/lng/odp.git/blob/HEAD:/example/switch/odp_switch.c) allows to exercise netmap in multiqueue mode. Number of Rx queues actually configured depends on CPU count and number of network interfaces supplied as parameters to the example binary.

I can't retest it, because I don't have the test environment anymore.

PR #95, opened 2015-10-07
Initialize time struct in receiver_body so code will compile.

john-gress:


john-gress:
Fixed by identical code added by Giuseppe Lettieri.

PR #97, opened 2015-10-08
fix printfs on 32 bit systems

skiingyac:
On a 32 bit system, the size_t types are not %lx but just %x.  But on a 64 bit system, the %lx is needed.

The fix is to use %zx where the z modifier automatically adjusts to match the 32/64 bit word size


vmaffione:
This should be rebased, unfortunately. However, I've already fixed this, you will see it upstream soon.

skiingyac:
Ok great!

Issue #106, opened 2015-10-22
NS_RMOREFRAG not set in RX with IXGBE


tbarbette:
Hi,

With LRO enabled, my Intel X520 NIC (ixgbe driver) is producing large packets that Netmap seems to split accross multiple buffers (well, it's the NIC which does it). But the NS_MOREFRAG is not set and I have no way to know that some packets are continuing on the next buffer.

sudo ethtool -K eth2 lro off
avoid the problem...

Thanks,
Tom

giuseppelettieri:
Hi Tom,

may we close this issue thanks to your solution which is now in master?

tbarbette:
Yes ! Thanks !

PR #107, opened 2015-10-22
Support NS_MOREFRAG with IXGBE

tbarbette:
Solve issue #106 . It was easier than I thought...

IXGBE driver now set/understand NS_MOREFRAG, allowing to use scatter gather I/O.


tbarbette:
Merged.

Issue #111, opened 2015-11-13
compat_ioctl for 32bit application


BASM:
For running 32bit a application on the 64bit kernel device driver (file_operations for ioctl) must have compat_ioctl vector. In the ioctl all data must use same struct on 32bit and 64bit struct.

Netmap have a good 'struct nmreq' and all other mmaped structs, but netmap do not init compat_ioctl vector (Why?). Therefore 32bit netmap is not works on the 64bit kernel.

The patch adds support for 32 bit applications, but I have not tested it well:

``` C
diff --git a/LINUX/netmap_linux.c b/LINUX/netmap_linux.c
index a9ca2dd..e225bb1 100644
--- a/LINUX/netmap_linux.c
+++ b/LINUX/netmap_linux.c
@@ -747,6 +747,9 @@ static struct file_operations netmap_fops = {
     .open = linux_netmap_open,
     .mmap = linux_netmap_mmap,
     LIN_IOCTL_NAME = linux_netmap_ioctl,
+#ifdef NETMAP_LINUX_HAVE_UNLOCKED_IOCTL
+    .compat_ioctl = linux_netmap_ioctl,
+#endif
     .poll = linux_netmap_poll,
     .release = linux_netmap_release,
 };
```

BASM:
I found two problems, its ssize_t and timvel types.

Struct netmap_ring has ssize_t on the ring_ofs param, and struct netmap_ring have struct timeval  ts.
If it fix, 32bit application on the 64bit kernel netmap works fine.

But I continue testing....

JamesZHuang:
Yes, the patch seems working, tested netmap-master(2016.9.16) on on CentOS 6.5.

vmaffione:
I provided the implementation of the compat_ioctl method. Can you test it?

PR #112, opened 2015-11-17
Supporting 64bit netmap API v11 in a 32bit applicaton.

BASM:
For build application you must add -DNETMAP_V11_64BIT option to the
CFLAGS env.

issue  #111


vmaffione:
This would need to be rebased

PR #116, opened 2015-12-11
linux dkms package

gureedo:
DKMS package for Debian-based Linux distributions.
This package can install netmap and nic drivers automatically and rebuild them on every linux kernel update.


vmaffione:
Thanks, we will try it out.

vmaffione:
thanks, merged in our internal repository, will be published soon

PR #129, opened 2016-01-14
Adjust netmap linux driver naming

binBASH:


vmaffione:
Hi,
Thank you, but the file you are touching are mostly outdated, and have to go away.
For instance, the real Archlinux PKGBUILD for netmap is available here:
https://aur.archlinux.org/packages/netmap/

PR #130, opened 2016-01-15
NETMAP_BDG_LIST bug

emctoo:
When listing all bridge/port with `vale-ctl`, the port column is not the port number, it's the index of the port number in the **bdg_port_index** in the kernel. Port number is not written into **nr_arg2** correctly in kernel code.


vmaffione:
this has been fixed upstream

PR #145, opened 2016-02-19
netmap_mem_global_finalize: fix bug triggered in error case

kogdenko:
If netmap_mem_global_config function failed and set nmd->lasterr
then nmd->active will be decremented.


vmaffione:
Thanks, merged. Will be published soon.

PR #159, opened 2016-03-29
added cksum in update_address function

victordetoni:
hi, I couldn't use the original cksum functions in update_address function, they were generating inconsistent cksum. So I added two new cksum functions and update_address is working fine when you use random function to send pkts.

Follow the full request for your analyse.


sbarber1968:
Can you please ensure you add the 
#ifndef NO_PCAP around the appropiate sections and remove the 'D' debug statements

Verified fix on my testbed

victordetoni:
hi,

the patch only created checksum functions and related them on
update_address function.

PR #164, opened 2016-04-15
rx.py a simple python receiver

abunimeh:
added an example UDP receiver using the NetmapDesc Python netmap class.

### DIFF
--- a/extra/python/rx.py
+++ b/extra/python/rx.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python
+# Description: Simple netmap rx example
+# Author: Faisal T. Abu-Nimeh (abunimeh@)
+
+import netmap
+import select
+
+IF_NAME = 'em0'
+
+d=netmap.NetmapDesc('netmap:' + IF_NAME)
+rxr = d.receive_rings[0] # single ring example
+# use poll
+poller = select.poll()
+poller.register(d.getfd(), select.POLLIN)
+
+while True:
+    ready_list = poller.poll() # wait for data
+    if len(ready_list) == 0:
+        print("Timeout occurred")
+        break;
+
+    if not rxr.empty(): # only process nonempty ring
+        i = rxr.cur
+
+        flen = rxr.slots[i].len
+        if flen > rxr.nr_buf_size: continue # drop large packets
+        fbuf = rxr.slots[i].buf[:flen] # this will fail if flen > nr_buf_size, i.e., nr_buf_size must be >= mtu
+
+        # drop everything except the following:
+        # if frame is an IP_pkt and UDP_pkt and and UDP_DST_PORT is 9956 
+        if fbuf[12:14] == '\x08\x00' and fbuf[23] == '\x11' and fbuf[36:38] == '\x26\xe4':
+            print "udp/ip pkt on my port"
+            print "netmap len=" + str(flen)
+            print "actual len=" + str(len(fbuf))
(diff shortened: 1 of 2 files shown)


vmaffione:
Thanks, but this should be a little more generic.

- Use argparse to get the interface name (I've done the same tx.py, still to be published), similarly to what is done by pktman.py.
- Use argparse to get the port number to select
- You assume there is no VLAN tag, so either you support it or you state it at program start/in the help.
- The name rx.py is not convenient, since this is not the receive equivalent of tx.py. We may call this udprecv.py.

Other comments inline

extra/python/rx.py
```diff
@@ -0,0 +1,39 @@
...
+    if len(ready_list) == 0:
+        print("Timeout occurred")
+        break;
+
+    if not rxr.empty(): # only process nonempty ring
+        i = rxr.cur
+
+        flen = rxr.slots[i].len
+        if flen > rxr.nr_buf_size: continue # drop large packets
+        fbuf = rxr.slots[i].buf[:flen] # this will fail if flen > nr_buf_size, i.e., nr_buf_size must be >= mtu
```
vmaffione: It cannot happen that flen > nr_buf_size you can drop these two instructions

extra/python/rx.py
```diff
@@ -0,0 +1,39 @@
...
+
+        # drop everything except the following:
+        # if frame is an IP_pkt and UDP_pkt and and UDP_DST_PORT is 9956 
+        if fbuf[12:14] == '\x08\x00' and fbuf[23] == '\x11' and fbuf[36:38] == '\x26\xe4':
+            print "udp/ip pkt on my port"
+            print "netmap len=" + str(flen)
+            print "actual len=" + str(len(fbuf))
+            print "first=%02x" % ord(fbuf[0])
+            print "last=%02x" % ord(fbuf[-1])
```
vmaffione: You should not add per-packet prints, since packets are expected to arrive at high rates. You can rate-limit the prints, or print statistics at the end

extra/python/rx.py
abunimeh: It can.

This actually occured to me that's why I committed this example.
I am running FreeBSD 10.3-RELEASE 
with the default sysctl dev.netmap.buf_size 
my ifconfig mtu was 9000 and my ethernet frame payload is around 8944.
flen was equal to the payload, i.e., 8944
len(fbuf) was equal to dev.netmap.buf_size (thas around 2048)
There were no flags indicating that I have fragments and the remaining values were in rxr.slots[i+1], rxr.slots[i+2], etc.

extra/python/rx.py
abunimeh: That's correct.

I just wanted to show how to view the received packet based on a testbed where the client is just sending single udp packets for testing.

extra/python/rx.py
vmaffione: Ok, if this happens then it is a bug, probably due to the fact that the netmap support for the adapter you are using (em?) does not check the EOP (end of packet) flag.
What happens with the slots immediately after the one you drop? are they complete packets or just fragments (associated to the dropped one)?

abunimeh: Yes em.

The slots immediately after the 1st one contain the remaining of the packet. However, each slot takes only 2048. If I recall correctly, If I don't drop the packet, I was able to recover it from the next 4 slots.

vmaffione:
Closing for now

PR #179, opened 2016-05-18
linux-pktgen.sh is actually bash, not posix sh

rbtcollins:


vmaffione:
Applied internally, thanks

PR #188, opened 2016-05-28
Also support setting the mode in linux pkg-gen.

rbtcollins:
This was useful in my testing to reproduce the panic.


vmaffione:
Merged in our internal repo, thanks!

PR #191, opened 2016-06-03
Generic adaptr release cleanups

rbtcollins:


rbtcollins:
Please review the generic_netmap_dtor change closely; I'm fairly sure /something/ like that is needed, but it doesn't solve my problem itself, and I don't know the surrounding logic well enough yet to know if this is sufficient - or correct - e.g. should we set na->ifp unconditionally still, or should it be guarded by the prev_na check too?

vmaffione:
I will try to reproduce the bug now, but I can tell for sure that the inverted check you propose is wrong. The netmap_adapter_put(prev_na) matches the one in generic_netmap_attach(), and it is needed.

I'm pretty sure this is not connected to our bug. The problem is not that we do or do not release the netmap adapter struct.

rbtcollins:
Ok, I'l close this, thanks for the review.

PR #197, opened 2016-06-08
unset the zombie flag on NETDEV_REGISTER (linux) and ifnet_arrival_event (freebsd)

tootoonchian:
Since commit d8d2569ba01bea1cc5497518125a63370f4551f5, netmap did not work with LXC containers. That is because, to move the device to a new namespace, it is unregistered first. Unregistering a netmap interface leaves it in a zombie state, which renders it unusable.

This patch adds listeners for the NETDEV_REGISTER (linux) and ifnet_arrival_event (freebsd) events and unsets the zombie flag in their handlers if needed.

Please note, I haven't tried the FreeBSD patch. Just tried to mirror what I did for the Linux one.

### DIFF
--- a/LINUX/netmap_linux.c
+++ b/LINUX/netmap_linux.c
@@ -81,6 +81,9 @@ linux_netmap_notifier_cb(struct notifier_block *b,
 
 	/* linux calls us while holding rtnl_lock() */
 	switch (val) {
+	case NETDEV_REGISTER:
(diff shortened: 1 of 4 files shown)


giuseppelettieri:
This looks good, thanks! We are going to test it and then merge.

vmaffione:
Merged, thanks!

PR #198, opened 2016-06-09
Fix #196 - document tcpdump behaviour with generic adapters.

### LINKED ISSUE #196: generic adapters forward all received traffic to host
I have a small test program which forwards all frames except gre frames from eth0 to eth0^, but when its running I still see gre packets in tcpdump; if I disable the filtering, I see each incoming gre packet twice.

rbtcollins:


share/man/man4/netmap.4
```diff
@@ -763,7 +763,10 @@ NICs without native support can still be used in
 mode through emulation. Performance is inferior to native netmap
 mode but still significantly higher than sockets, and approaching
 that of in-kernel solutions such as Linux's
-.Xr pktgen .
+.Xr pktgen . Programs using taps such as tcpdump will see packets before they
+are diverted by netmap when emulation is in use. When such packets are
+forwarded to the host stack they show up on the tap a second time, but the host
+OS stack will only see such forwarded packets once.
```
vmaffione: We should say "could see" rather than "will see". I agree that you can stress that this is unwanted, just an artifact of the implementation.

We should definitely remove the sentence "When such pakcets are forwarded".. because it's confusing, and it is not something related to the emulated mode. Packets are forwarded by your application (e.g. bridge), this is not part of the generic/emulated. Moreover, the fact that you see packets with TCP dump (the second time, let's say) it's due to the normal functioning of the host RX ring, which is already documented somewhere else.

rbtcollins:
Ok, so will see -> could see; add some prose about artifact.

W.r.t. the forwarded sentence - I understand the confusion, I would like to preserve the expansion of what is otherwise only implied - that you can dump the same packet twice if the netmap application chooses to send it unaltered to the host ring. Would you be ok with that if I rephrase to avoid 'forwarded' ?

vmaffione:
I'm not sure it's a good idea, but you could try to reformulate with "If", not "when".
Let's also replace "program using taps" with "packets sniffer programs".

rbtcollins:
reformulated, what do you think?

vmaffione:
slightly rearranged and merged, thanks!

PR #203, opened 2016-08-04
Fix possible memory leak if netmap_get_na fails

JaseFace:


vmaffione:
Thanks, merged in our internal server. Will be published soon.

PR #211, opened 2016-08-19
Switch number of iteration to an uint64_t

ocochard:
Current packets counter should be at minimum an u_int and not an int.
On 10/40Gbps links or bigger we need to generate large number of packets, and the current code generate a Segmentation fault, then switch this variable to uint64_t.

Before this patch:
```
[root@hp]/tmp# ./pkt-gen -i ncxl0 -f tx -n 80000000000
170.336326 main [2235] interface is ncxl0
170.336341 main [2355] running on 1 cpus (have 8)
170.336391 extract_ip_range [364] range is 10.0.0.1:0 to 10.0.0.1:0
170.336398 extract_ip_range [364] range is 10.1.0.1:0 to 10.1.0.1:0
170.351471 main [2456] mapped 334980KB at 0x801800000
170.357651 get_vnet_hdr_len [792] Unable to get virtio-net header length
Sending on netmap:ncxl0: 2 queues, 1 threads and 1 cpus.
10.0.0.1 -> 10.1.0.1 (00:00:00:00:00:00 -> ff:ff:ff:ff:ff:ff)
170.357667 main [2553] Sending 512 packets every  0.000000000 s
170.357669 main [2555] Wait 2 secs for phy reset
172.463645 main [2557] Ready...
172.463711 sender_body [1174] start, fd 3 main_fd 3
Segmentation fault
```

### DIFF
--- a/examples/pkt-gen.c
+++ b/examples/pkt-gen.c
@@ -233,7 +233,7 @@ struct glob_arg {
 	int pkt_size;
 	int burst;
 	int forever;
-	int npackets;	/* total packets to send */
+	uint64_t npackets;	/* total packets to send */
 	int frags;	/* fragments per packet */
 	int nthreads;
 	int cpus;	/* cpus used for running */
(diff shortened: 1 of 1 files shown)


vmaffione:
Hi,
  Thanks, merged in our internal repo, will be published soon.

PR #225, opened 2016-09-23
add get_ethtool_ops_ext in nm_os_generic_find_num_queues function

alandtsang:
In the linux-2.6.32 kernel, the struct ethtool_ops has no get_channels function, get_channels is in another ethtool extension struct ethtool_ops_ext.


vmaffione:
This breaks the compilation of all the netmap versions

PR #233, opened 2016-10-08
Fix memory leak

alandtsang:
Fix memory leak in function sender_body and main_thread

### DIFF
--- a/examples/pkt-gen.c
+++ b/examples/pkt-gen.c
@@ -1317,6 +1317,7 @@ sender_body(void *data)
 quit:
 	/* reset the ``used`` flag. */
 	targ->used = 0;
+  free(targ->nmd);
 
 	return (NULL);
 }
@@ -2094,6 +2095,7 @@ main_thread(struct glob_arg *g)
 			tic = timespec2val(&targs[i].tic);
 		if (!timerisset(&toc) || timespec_ge(&targs[i].toc, &t_toc))
 			toc = timespec2val(&targs[i].toc);
+    free(&targs[i]);
 	}
 
 	/* print output. */


examples/pkt-gen.c
```diff
@@ -2094,6 +2095,7 @@ main_thread(struct glob_arg *g)
 			tic = timespec2val(&targs[i].tic);
 		if (!timerisset(&toc) || timespec_ge(&targs[i].toc, &t_toc))
 			toc = timespec2val(&targs[i].toc);
+    free(&targs[i]);
```
vmaffione: wrong, targs are allocated all together, not cell by cell

examples/pkt-gen.c
```diff
@@ -1317,6 +1317,7 @@ sender_body(void *data)
 quit:
 	/* reset the ``used`` flag. */
 	targ->used = 0;
+  free(targ->nmd);
```
vmaffione: wrong, nm_close() must be used

vmaffione:
Problems fixed in the correct way, it will be published in a while

examples/pkt-gen.c
vmaffione: yes

examples/pkt-gen.c
vmaffione: yes --> free(targs)

examples/pkt-gen.c
alandtsang: OK

examples/pkt-gen.c
alandtsang: Thanks, I got it

PR #252, opened 2016-11-20
Add an option for disabling normalized number: non-fixed unit is complex to parse

ocochard:
The new behavior of pkt-gen that automatically normalize unit (by converting to Kpps, Mpps, Gpps, Tpps) is a good idea for a human reader, but a bad one for my automated benchmark tests:
It's lot's more difficult to parse pkt-gen value now.
Then here is an option for disabling this unit normalization.


vmaffione:
Thanks, but I think that there is a better way to support this, without the replication of the printf code.
We can pass "normalize" as an argument to the norm() function and decide what to do inside there. Could you try this approach?

giuseppelettieri:
I think we can close this, since equivalent functionality has already been merged.

PR #278, opened 2017-02-23
pkt-gen: do not use sizeof(char*) with strncmp

funman:
ifa_name is a char* on Linux and FreeBSD


vmaffione:
Good catch, thanks. To stay safe, would you mind changing the code so that we still use strncmp(), but passing IFNAMSIZ as a third argument?

funman:
I did just that

vmaffione:
applied, thanks

PR #280, opened 2017-02-28
bump ring pool size

funman:
sizeof(struct netmap_slot) is 4+2+2+8 == 16
Assuming PAGE_SIZE is 4096, 17*4096 is enough to allocate
4096 slot entries + sizeof(struct netmap_ring)

This makes it possible to use 4096 TX buffers with the ixgbe driver

### DIFF
--- a/sys/dev/netmap/netmap_mem2.c
+++ b/sys/dev/netmap/netmap_mem2.c
@@ -418,7 +418,7 @@ struct netmap_mem_d nm_mem = {	/* Our memory allocator. */
 			.num  = 100,
 		},
 		[NETMAP_RING_POOL] = {
-			.size = 9*PAGE_SIZE,
+			.size = 17*PAGE_SIZE,
 			.num  = 200,
 		},
 		[NETMAP_BUF_POOL] = {


giuseppelettieri:
Why don't you just increase the ring_size via the dev.netmap.ring_size sysctl (FreeBSD) or /sys/module/netmap/parameters/ring_size (linux)?

funman:
I need to try that, I found the size definition by going backwards from the kernel error message.

Perhaps I should replace that change by documenting ring_size ;)

vmaffione:
Yes, indeed documentation can be improved.
Here is an fragment of a ML discussion on freebsd-net, if useful:

"
[...]
In any case, buf_size, ring_size and if_size are the sizes in bytes of each buffer, ring and netmap_if (control data structure), respectively.
So the maximum amount of slots for each ring is (ring_size-sizeof(struct netmap_ring))/16, as 16 is the size in bytes of struct netmap_slot.

On the other side, buf_num, ring_num and if_num are the total number of buffers, rings and netmap_if objects in each netmap memory area (aka "allocator").
By default there is a single memory area used by all the hardware NICs and a separate memory area for each VALE port.
This is already configurable, however.
[...]
"

funman:
Closing due to ring_size already doing this job

PR #284, opened 2017-03-17
Fix 2-tuple fallback hashing in lb.

sethhall:
Apparently I broke the fallback (non tcp/udp) hashing mechism for IPv4 for some reason.  This fixes it.


giuseppelettieri:
merged internally and now published. Thanks!

PR #298, opened 2017-04-16
Fix typo in Linux Configure Script

gspivey:
Linux Configure script has a typo in the help text.
"destir" should be "destdir"


vmaffione:
Thanks

Issue #303, opened 2017-05-12
False negative message


spinenkoia:
Good afternoon.
I found the following messages in kernel.log when loading (and unloading) the kernel module netmap.
```
Kernel: 307.695732 [3385] netmap_init run mknod / dev / netmap c 10 54 # error 0
Kernel: netmap: loaded module
Kernel: netmap: unloaded module.
```
But they are all with the level of ERROR, although it should be INFO.
Looked at the source code, there's that.
```
printf ("netmap: loaded module \ n");
return (0);
```
A printf, it is defined like this.
`#define printf (fmt, arg ...) printk (KERN_ERR fmt, ## arg)`

vmaffione:
Yes, for portability across FreeBSD and Linux we do not preserve the log level.
That's not clean, I know. We should have two separate sets of macros, one for errors and one for info, and change all the places in the code where the macros are used (a lot).

socketpair:
```C
#ifdef FREEBSD
#define log (level, message ...) ???
#else
#define log (level, arg ...) printk (level message, ##arg)
#endif

// And after that in code :

log(KERN_ERR, "message");
log(KERN_INFO, "message");
```

Why not ?

vmaffione:
Yes, that's exactly what I said, but you need to change all the usages of D() and RD() in the code.

socketpair:
Well, I have editted comment and provide quicker fix

spinenkoia:
Here, I created pullrequest https://github.com/luigirizzo/netmap/pull/304

PR #305, opened 2017-05-15
Added log levels for printf function as in linux.

spinenkoia:
https://github.com/luigirizzo/netmap/issues/303

### DIFF
--- a/LINUX/bsd_glue.h
+++ b/LINUX/bsd_glue.h
@@ -58,7 +58,6 @@
 #include <linux/io.h>	// virt_to_phys
 #include <linux/hrtimer.h>
 
-#define printf(fmt, arg...)	printk(KERN_ERR fmt, ##arg)
 #define KASSERT(a, b)		BUG_ON(!(a))
 
 /*----- support for compiling on older versions of linux -----*/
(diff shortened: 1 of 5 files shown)


vmaffione:
A patch in this direction is very welcome, thanks.
But this is not the approach we are looking for, we don't want to inherit Linux KERN_** macros.
We would like to have two macros, one to print errors and one to print info, e.g.,
````
  nm_prerr("error message %d", xxx)
  nm_prinf("info message %d", xxx)
````
On Linux nm_prerr maps to printk(KERN_ERR) and nm_prinf maps to printk(KERN_INFO).
On FreeBSD both can map to printf().
Could you please align your patch in this direction?

spinenkoia:
Yes OK.
Did I understand correctly that instead of a single function, do you need to use nm_prrr and nm_printer?

vmaffione:
Yes, netmap code will use one function or the other depending on the specific needs.

spinenkoia:
@vmaffione Look please that it turned out, is such a decision satisfied?

vmaffione:
It looks good, but we need to check if compilation is ok on FreeBSD. Have you checked it by any chance?

spinenkoia:
FreeBSD has not yet been tested, only in Linux. In the near future I will check.

vmaffione:
Thanks. I've checked on FreeBSD, it is ok.
Next step would be to get rid of the D,ND and RD macros and use nm_prinf and nm_prerr everywhere.

spinenkoia:
Yes, I think I will do it. It would also be useful for me.

vmaffione:
Sounds good! Actually, we can leave ND macros (they result into nothing), and we need another nm_prlim() macro to replace RD.

spinenkoia:
Okay, I'll keep that in mind.

PR #315, opened 2017-06-13
Silence a (fatal) warning in FreeBSD when building one of the platfor…

loos-br:
…ms that still requires GCC (powerpc, powerpc64 and sparc64).


vmaffione:
Ok thanks

PR #321, opened 2017-06-26
i40e: Set NS_MOREFRAG slot flag if frame is incomplete

desbma:
For Ethernet frames of size larger than 2048 bytes, the ring for this driver always gives buffers of 2048 bytes or smaller chunks, even if the MTU is configured for > 9000 bytes, but it exposes a flag to inform the driver that more fragments are available for this frame.

This patch propagates the "more fragment" flag, from the NIC ring to the Netmap ring.
A similar check is done [in the mainline Linux i40e driver](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/drivers/net/ethernet/intel/i40e/i40e_txrx.c?h=v4.11.7#n1727).

Without this fix, when a 9014 byte frame is captured on the interface, a Netmap reader would receive 5 chunks of sizes : 2048 + 2048 + 2048 + 2048 +  822, without a way to tell if they belong to the same frame or not.

I have tested the change on Intel XL710 10Gbe card.

### DIFF
--- a/LINUX/i40e_netmap_linux.h
+++ b/LINUX/i40e_netmap_linux.h
@@ -455,6 +455,7 @@ i40e_netmap_rxsync(struct netmap_kring *kring, int flags)
 	if (netmap_no_pendintr || force_update) {
 		int crclen = ix_crcstrip ? 0 : 4;
 		uint16_t slot_flags = kring->nkr_slot_flags;
+		uint16_t curr_slot_flag;
 
 		nic_i = rxr->next_to_clean; // or also k2n(kring->nr_hwtail)
 		nm_i = netmap_idx_n2k(kring, nic_i);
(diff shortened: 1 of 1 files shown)


LINUX/i40e_netmap_linux.h
```diff
@@ -470,7 +471,13 @@ i40e_netmap_rxsync(struct netmap_kring *kring, int flags)
 			}
 			ring->slot[nm_i].len = ((qword & I40E_RXD_QW1_LENGTH_PBUF_MASK)
 			    >> I40E_RXD_QW1_LENGTH_PBUF_SHIFT) - crclen;
-			ring->slot[nm_i].flags = slot_flags;
+
+			curr_slot_flag = slot_flags;
+			if (unlikely((staterr & (1<<I40E_RX_DESC_STATUS_EOF_SHIFT)) == 0 )) {
+				curr_slot_flag |= NS_MOREFRAG;
```
giuseppelettieri: Why the need for curr_slot_flags? Can't you just replace the above with `ring->slot[nm_if].flags |= NS_MOREFRAG`?

desbma: For the same reason there is a `slot_flags` variable whose only purpose is to hold a copy of `kring->nkr_slot_flags`.

I find it more readable, and it also avoids a second `ring->slot[nm_i].flags` **triple** indirection which is bad for performance (although any decent compiler would probably optimize it to the same code).

giuseppelettieri: of course the compiler would optimize it: my point is only about readability, since the new variable is distracting. Anyway, this is not really important. We can merge the the patch as is, if you like.

desbma: Well, it's for you to decide. :)  
If you can merge it as is, it's great.
If you find the temporary variable distracting, I can remove it, I don't care that much.

giuseppelettieri: let's keep it, then. Thanks!

PR #324, opened 2017-06-29
Use htons to initialize packet's ip_len

borislavmatvey:
In the expected context the use of htons is more appropriate. htons and
ntohs are actually implemented using the same function on x86, so there
is no actual functional problem.


vmaffione:
Sure, thanks.

PR #328, opened 2017-07-03
netmap_do_regif: Reverse error path vs normal path

borislavmatvey:
The error exit path should reverse all operations in backward order of
the normal path. Till this commit if netmap_mem_if_new fails then
netmap_krings_put won't be called and the usage count will leak.


giuseppelettieri:
Thanks!

PR #329, opened 2017-07-03
netmap_ioctl: Fix ifnet reference leak on failure.

borislavmatvey:
The redefinition of ifp variable in the registration path was shadowing
a definition of the same variable at the start of netmap_ioctl function
and in case of netmap_do_regif failure the error path was trying to
unref the outer variable which contains NULL, while the ifnet object was
pointed by the inner ifp variable and as it went out of scope the ifnet
object will have one extra reference.

### DIFF
--- a/sys/dev/netmap/netmap.c
+++ b/sys/dev/netmap/netmap.c
@@ -2312,7 +2312,6 @@ netmap_ioctl(struct netmap_priv_d *priv, u_long cmd, caddr_t data, struct thread
 		NMG_LOCK();
 		do {
 			u_int memflags;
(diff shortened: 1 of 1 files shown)


borislavmatvey:
This modification will fix the leak, but I'm not sure this is the best change. I saw that this variable is redefined in a couple of places in the ioctl function and similar problem could happen in the future.

Any notes or suggestions are kindly welcome.

desbma:
Maybe compiling with `-Wshadow` or even `-Werror=shadow` would help catch these kind of bugs in the future?

Issue #322, opened 2017-06-28
Kernel panic on FreeBSD-CURRENT with loopback


larseggert:
Trying to use netmap over loopback crashes the kernel under FreeBSD-CURRENT. Here is the console log:
```
414.499376 [ 389] nm_os_generic_find_num_desc called, in tx 1024 rx 1024
414.574290 [ 397] nm_os_generic_find_num_queues called, in txq 0 rxq 0
414.649158 [1252] generic_netmap_attach     Created generic NA 0xfffff80004643400 (prev 0)
414.915872 [ 448] generic_netmap_register   Generic adapter 0xfffff80004643400 goes on
415.005343 [ 495] generic_netmap_register   RX ring 0 of generic adapter 0xfffff80004643400 goes on
415.110371 [ 502] generic_netmap_register   TX ring 0 of generic adapter 0xfffff80004643400 goes on


Fatal trap 12: page fault while in kernel mode
cpuid = 1; apic id = 01
fault virtual address	= 0x0
fault code		= supervisor read instruction, page not present
instruction pointer	= 0x20:0x0
stack pointer	        = 0x28:0xfffffe0476ef11e8
frame pointer	        = 0x28:0xfffffe0476ef1220
code segment		= base 0x0, limit 0xfffff, type 0x1b
			= DPL 0, pres 1, long 1, def32 0, gran 1
processor eflags	= interrupt enabled, resume, IOPL = 0
current process		= 2025 (test_warp)
[ thread pid 2025 tid 100180 ]
Stopped at      0
db> bt
Tracing pid 2025 tid 100180 td 0xfffff8001b4b1a00
??() at 0
drbr_enqueue_drv() at drbr_enqueue_drv+0x54c/frame 0xfffffe0476ef1220
nm_os_generic_xmit_frame() at nm_os_generic_xmit_frame+0x48/frame 0xfffffe0476ef1230
generic_netmap_attach() at generic_netmap_attach+0xe79/frame 0xfffffe0476ef1310
netmap_ioctl() at netmap_ioctl+0x752/frame 0xfffffe0476ef16f0
nm_os_selrecord() at nm_os_selrecord+0x1de/frame 0xfffffe0476ef1720
dev2udev() at dev2udev+0x2b40/frame 0xfffffe0476ef1770
VOP_IOCTL_APV() at VOP_IOCTL_APV+0x89/frame 0xfffffe0476ef17a0
[...]

PR #338, opened 2017-07-21
Do not allow the use of the loopback interface in netmap.

loos-br:
The generic support in netmap send the packets using if_transmit() and the
loopback do not support packets coming from if_transmit()/if_start().

This avoids the use of the loopback interface and the subsequent crash that
happens when the application send packets to the loopback interface.

Details in:	https://github.com/luigirizzo/netmap/issues/322
Obtained from:	FreeBSD


loos-br:
Do not allow the use of the loopback interface in netmap. *only for FreeBSD*

PR #349, opened 2017-08-18
Respect dkms $kernelver

funman:
Fix kernel upgrades when the module is built for kernel different from the one running


funman:
Ping

PR #351, opened 2017-08-25
Various pkt-gen fixes

stephen-hurd:
- Add a -N option to not normalize units.  Useful for automated tools.
  (From olivier@ here:
   https://github.com/ocochard/BSDRP/tree/master/BSDRPcur/patches)
- Or in -o option so command-line order of -I, -X, -r, -z, -Z, and -A
  doesn't matter.
- Avoid unaligned access, some new compilers will warn on it, and pkt-gen
  uses -Werror -Wall.
- Document all supported options in usage() (even -m).
- Remove documentation of unsupported -r and -t options
- Align tabs in usage()

PR #352, opened 2017-08-25
pkt-gen: update help

funman:
-t options never existed, and -r means something else


stephen-hurd:
Heh, see #351, I just did a pull request with a massive update to usage()

funman:
Ah should have checked before ;)

Also I see you left the option here for future implementation, that's nice

Closing this one

PR #353, opened 2017-08-28
Check if carrier present instead of running.

borislavmatvey:
Replace the check whether interface is running with a check whether
there is actual carrier, otherwise packets will be reported as sent
although there is no actual link on the interface.

Tested by sending packets with:
```
pkt-gen -i ethX -f tx -n 0 -l 1500
```

Pkt-gen shows that it is sending packets even after the cable was
physically unplugged from the interface.

The check in ixgbe is the same in txsync handler.

### DIFF
--- a/LINUX/i40e_netmap_linux.h
+++ b/LINUX/i40e_netmap_linux.h
@@ -283,7 +283,7 @@ i40e_netmap_txsync(struct netmap_kring *kring, int flags)
 	struct i40e_vsi *vsi = np->vsi;
 	struct i40e_ring *txr;
 
-	if (!netif_running(ifp))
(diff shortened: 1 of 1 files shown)


borislavmatvey:
Fixes #366

vmaffione:
Maybe we should check for both? @giuseppelettieri

borislavmatvey:
Just to mention I tested this works as expected when plugged in/out as it is in the PR - and this check done now is the same as in done in the same function for the `ixgbe` driver.

giuseppelettieri:
I am closing this, thanks for the patch!

borislavmatvey:
Thanks!

PR #377, opened 2017-10-11
Fixed lut_entry has no paddr member issue

nanjj:
The error: 'struct lut_entry' has no member named 'paddr'


vmaffione:
Thanks

giuseppelettieri:
Thanks for trying to compile this, but the fix is different and will be provided in upcoming commits. Your solution is writing through a NULL pointer.

PR #391, opened 2017-11-02
mem: fix non-const casting of nr_buf_size

carlgsmith:
One some platforms this previously resulted in
nr_buf_size being initialised to random values
as only 16-bits of the 32 bit value was set.
The symtoms were that buffer offsets were
incorrectly calculated and invalid memory
accesses occurred.


vmaffione:
good catch, thanks

Issue #409, opened 2017-12-13
Missing dma synchronization from device


dusancerhaty:
Packets received from NIC are not consistent with packets sent to device. It happens quite often. It's caused by missing dma_buffers/caches sync, like DMA-API documentation recommends for streaming DMA mappings, using dma_sync_single_for_*(). I attached a patch solving the issue for RX rings only, as this is all I need now. There's a need to fix it for TX rings also.

Could you take a look please?

Thanks

[0002-mem-sync-dma-rx-buffers-from-device.txt](https://github.com/luigirizzo/netmap/files/1557023/0002-mem-sync-dma-rx-buffers-from-device.txt)

giuseppelettieri:
Thanks. As I mentioned elsewhere, DMA-mapping was skipped during the port to Linux, and has been left in that state for a long time, since it apparently did not make any difference on Intel-based architectures. Are you experiencing these cache problems on ARM?

Either way, please prepare a pull request for your patch.

dusancerhaty:
Yes, this issue emerged on ARM. But after I've read your post, I tested it on intel-based machine without syncing and the problem didn't appear. In spite of this I wouldn't rely on this test result. I didn't find any mention about syncing not being necessary on x86, nor another architecture. Also DMA-API doc recommends to sync. So I leave the patch as is, e.g. it'll be supported for all archs, it has no affect on performance.

giuseppelettieri:
Thanks for testing. No, it is generally not needed on Intel, and that is why we could avoid supporting it for so long. Anyway, I agree that we should do right thing and follow the API. In fact, I was not suggesting to drop the patch. To the contrary, I would like to merge it with proper credit and apply the change to the other drivers and to the TX path.

Up until recently we did not support ARM at all, so you are up for a tough ride :)

giuseppelettieri:
I am leaving this open since it needs to be done on the other drivers, too. Moreover, we need to do the right thing on the TX path also. We may need to some help to test that, though, since we do not own any ARM hardware.

dusancerhaty:
Now I'm trying to find out how to correctly handle syncing of TX buffers, so I can manage TX path if you wish.

Regarding the testing on ARM, do no hesitate to contact me whenever in the future.

Further, at the beginning of next year, I'm going to implement full support for IOMMU, 'cause I need to manage more NIC's ports on one machine what is not possible now if they belongs to different IOMMU groups.

Anyway, great job done on this project, saved me a lot time.

Issue #407, opened 2017-12-13
Unloading of DMA mappings is not correct


dusancerhaty:
Mappings created by netmap_load_map() are not unmapped correctly. The cycle, which maps paddrs using dma_map_single(), starts at index 0 and use p->_clustsize, but the cycle, which unloads them, starts at index 2 and use size of netmap buf size. It results in freeing of only a small portion of previously mapped area, thus leading to map another parts of memory on next interface registering, as those previously mapped are not usable. After couple of runs of my test utility I end up in kernel oop, as linux is not able to handle paging request. The problem is more significant when IOMMU is enabled.

I am trying to use netmap on NVIDIA's jetsontx2, Ubuntu 16.04, Linux 4.4.38, but I think it's common problem of any platform.

I attached a patch resolving this issue, of course I am not able to say whether it is corrent. Could you please take a look?

Thanks

[0001-mem-fix-unloading-of-dma-mappings.txt](https://github.com/luigirizzo/netmap/files/1555002/0001-mem-fix-unloading-of-dma-mappings.txt)

giuseppelettieri:
Thanks for catching this. Please prepare a pull request.

giuseppelettieri:
merged, I am closing this.

PR #410, opened 2017-12-14
Dma mapping fixes

dusancerhaty:
Resolves issuess #407 and #409.

### DIFF
--- a/sys/dev/netmap/netmap_kern.h
+++ b/sys/dev/netmap/netmap_kern.h
@@ -1632,6 +1632,8 @@ netmap_unload_map(struct netmap_adapter *na,
 		bus_dmamap_unload(tag, map);
 }
 
+#define netmap_sync_map(na, tag, map, sz, t)
+
 /* update the map when a buffer changes. */
 static inline void
 netmap_reload_map(struct netmap_adapter *na,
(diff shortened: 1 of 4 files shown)


vmaffione:
(we need to check that this does not break compilation on FreeBSD)

sys/dev/netmap/netmap_kern.h
```diff
@@ -1666,16 +1666,24 @@ netmap_load_map(struct netmap_adapter *na,
...
-	u_int sz = NETMAP_BUF_SIZE(na);
-
 	if (*map) {
 		dma_unmap_single(na->pdev, *map, sz,
 				 DMA_BIDIRECTIONAL);
 	}
 }
 
+static inline void
+netmap_sync_map(struct netmap_adapter *na,
```
giuseppelettieri: In anticipation for the TX path changes, may you please pass the direction as an additional ```enum txrx``` parameter?

dusancerhaty: I've added the parameter, but left function body untouched, 'cause I'm a bit confused about what dma_sync_single_for_*() call and DMA_* direction to use in case of NR_TX ring.

dusancerhaty:
I fixed FreeBSD compilation by declaring netmap_sync_map() as noop macro.

PR #411, opened 2017-12-15
Sync dma for tx path

dusancerhaty:
Syncing of DMA maps now works for TX path also. Integrated and tested for IGB, as I have NICs using this driver only.

Relates to issue #409.

### DIFF
--- a/LINUX/if_igb_netmap.h
+++ b/LINUX/if_igb_netmap.h
@@ -159,7 +159,7 @@ igb_netmap_txsync(struct netmap_kring *kring, int flags)
 				// netmap_reload_map(pdev, DMA_TO_DEVICE, old_paddr, addr);
 			}
 			slot->flags &= ~(NS_REPORT | NS_BUF_CHANGED);
-			netmap_sync_map(na, (bus_dma_tag_t) na->pdev, &paddr, NETMAP_BUF_SIZE(na), NR_TX);
+			netmap_sync_map(na, (bus_dma_tag_t) na->pdev, &paddr, len, NR_TX);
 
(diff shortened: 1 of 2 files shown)


dusancerhaty:
It has obvious performance impact also when not transmitting packets, only receiving. Do not merge, I try to rethink.

giuseppelettieri:
The performance impact is not obvious to me. May you elaborate a bit?

LINUX/if_igb_netmap.h
```diff
@@ -159,6 +159,7 @@ igb_netmap_txsync(struct netmap_kring *kring, int flags)
 				// netmap_reload_map(pdev, DMA_TO_DEVICE, old_paddr, addr);
 			}
 			slot->flags &= ~(NS_REPORT | NS_BUF_CHANGED);
+			netmap_sync_map(na, (bus_dma_tag_t) na->pdev, &paddr, NETMAP_BUF_SIZE(na), NR_TX);
```
giuseppelettieri: It probably better to pass the actual size of of the packet, rather than ```NETMAP_BUF_SIZE(na)```. This was also the case for the rx path.

dusancerhaty:
I tested it again intensively and the impact on performance didn't appear. I do not know what was the cause of performance impact before, but it seems it wasn't syncing of DMA buffers. I also changed the size passed to the netmap_sync_map() to the size of packet according to your suggestion.

giuseppelettieri:
I can merge, if you agree.

dusancerhaty:
Yes, I agree.

Issue #398, opened 2017-11-19
The max ports in a vale should be 254


nanjj:
Steps to reproduce the problem:
1. create and attach 252 ports in valea:
```
#bash
for ((i=0; i<252; i++)); do
    vale-ctl -n a$i
    vale-ctl -a valea:a$i
done
```
2. create and attach the 253 port
```
vale-ctl -n a252
vale-ctl -a valea:a252
```

=>
valea:a252: Cannot allocate memory

Why?

The max ports in a vale should be 254 according `NM_BDG_MAXPORTS` in netmap_kern.h.

giuseppelettieri:
Memory allocation may fail for other reasons. What is the error logged to dmesg?

snowcherry:
... ...
[  116.513050] 148.545852 [ 676] netmap_vi_create          returning nr_arg2 250
[  116.515913] 148.548714 [ 680] netmap_vi_create          created a248
[  116.520359] 148.553157 [2286] netmap_vp_create          nr_arg2 0
[  116.523191] 148.555987 [ 676] netmap_vi_create          returning nr_arg2 251
[  116.526044] 148.558839 [ 680] netmap_vi_create          created a249
[  116.530388] 148.563179 [2286] netmap_vp_create          nr_arg2 0
[  116.533194] 148.565985 [ 676] netmap_vi_create          returning nr_arg2 252
[  116.536046] 148.568836 [ 680] netmap_vi_create          created a250
[  116.540528] 148.573314 [2286] netmap_vp_create          nr_arg2 0
[  116.543352] 148.576136 [ 676] netmap_vi_create          returning nr_arg2 253
[  116.546207] 148.578990 [ 680] netmap_vi_create          created a251
[  173.993979] 205.992647 [ 757] netmap_get_bdg_na         bridge full 252, cannot create new port

nanjj:
[  116.546207] 148.578990 [ 680] netmap_vi_create          created a251
[  173.993979] 205.992647 [ 757] netmap_get_bdg_na         bridge full 252, cannot create new port

PR #421, opened 2017-12-26
Changed NM_BRIDGES to 256

nanjj:
Added GTAGS to .gitignore and fixed vale port max not 254 issue #398 

To NM_BRIDGES default to 256 because the original default value `8` 
is too small for even test, and `256` is OK for hypervisor usage.

### DIFF
--- a/sys/dev/netmap/netmap_vale.c
+++ b/sys/dev/netmap/netmap_vale.c
@@ -711,7 +711,6 @@ netmap_get_bdg_na(struct nmreq *nmr, struct netmap_adapter **na,
 	struct netmap_vp_adapter *vpna, *hostna = NULL;
 	struct nm_bridge *b;
 	int i, j, cand = -1, cand2 = -1;
-	int needed;
 
 	*na = NULL;     /* default return value */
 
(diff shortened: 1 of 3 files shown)


sys/dev/netmap/netmap_vale.c
```diff
@@ -752,8 +751,7 @@ netmap_get_bdg_na(struct nmreq *nmr, struct netmap_adapter **na,
 	if (!create)
 		return ENXIO;
 	/* yes we should, see if we have space to attach entries */
-	needed = 2; /* in some cases we only need 1 */
-	if (b->bdg_active_ports + needed >= NM_BDG_MAXPORTS) {
+	if (b->bdg_active_ports >= NM_BDG_MAXPORTS) {
```
giuseppelettieri: The previous check was too conservative: ```>=``` should have been ```>``` and ```needed``` should have been computed more precisely.

Your change, however, is wrong. There _are_ cases when we need two slots. See where ```cand2``` is actually used.

jolisexy: hello how are you me it is natacha I am unmarried Canadian here to make me friends here is my skype natacha dipou sorry for the inconvenience thank you and have a good day as you

PR #468, opened 2018-03-28
Update bsd_glue.h

wangbojing:
kernel 4.11.0 atomic convert to refcount.


vmaffione:
Why is this needed? Do you get a compilation failure?
In any case we have a macro `NETMAP_LINUX_HAVE_REFCOUNT_T`, we never check the versions on linux.

PR #476, opened 2018-04-17
Add support for Mellanox ConnectX-4 / ConnectX-5 cards (v3.3 mlx5 driver)

### LINKED ISSUE #210: Mellanox ConnectX-4 support
Hi,

I am wondering whether there is any netmap effort being put for Mellanox drivers.
I have a single port ConnectX4 100GbE card which uses mlx4-en and mlx4-core modules and I want to test how far I can go, in terms of Tx and Rx capabilities, with netmap.

Thanks in advance,
Georgios

andrewbonney:
Hi,
As previously discussed, please find attached patches for the Mellanox mlx5 Ethernet driver for use with ConnectX-4 and ConnectX-5 cards (Resolves #210).

As we build things a little differently internally I've refactored our internal code to suit the netmap codebase, following the same sort of method used to pull in external Intel drivers for patching (as opposed to patching the kernel source).

Whilst I am able to build this, I haven't yet tackled the necessary changes to LINUX/default-config.mak.in_ to enable downloading and patching of the relevant tarball. I'm hoping the details below will help with this, and I'm happy to make further changes as required to make it fit in.

Build process:
* Download tgz from http://www.mellanox.com/downloads/Drivers/mlnx-en-3.3-1.0.0.0.tgz
* Unpack, then unpack from the tgz inside SOURCES/
* Patch (patch --posix --quiet --force -p1 < patches/mellanox--mlx5--3.3;)
* Run scripts/mlnx_en_patch.sh --without-mlx4 before make (using '-j\<x\>' to speed up as necessary) - this deals with detecting and patching for the current kernel within the Mellanox driver.
* This builds three ko files but only mlx_compat.ko and mlx5_core.ko are required for the ethernet driver.

Proposed resulting additions to resultant drivers.mak
```
mlx5@conf := CONFIG_MLX5
mlx5@src := tar xf /project-mcuk/ap/ipp/andrewbo/netmap/LINUX/ext-drivers/mlnx-en-3.3-1.0.0.0.tgz && tar xf /project-mcuk/ap/ipp/andrewbo/netmap/LINUX/mlnx-en-3.3-1.0.0.0/SOURCES/mlnx-en_3.3.orig.tar.gz && ln -s mlnx-en-3.3 mlx5
mlx5@patch := patches/mellanox--mlx5--3.3
```

The patches are against Mellanox driver version 3.3-1, and have been tested against various Linux kernels from 3.13 upwards, although not all the way to the most recent one. Our recent builds have been tested mostly against kernel 4.4 using Ubuntu 16.04.

In time these could be ported onto the version of the mlx5 driver which is bundled with the Linux kernel, but there isn't an exact match for v3.3-1 of the driver within the kernel, so I've not done this as yet.

I hope this is useful, and please let me know of any further modifications which might be required in order to assist with merging it.

Thanks

### DIFF
--- a/LINUX/configure
+++ b/LINUX/configure
@@ -110,7 +110,7 @@ subsys enable generic
 
 # available drivers
 driver_avail="r8169.c virtio_net.c forcedeth.c veth.c \
-	e1000 e1000e igb ixgbe ixgbevf i40e vmxnet3"
+	e1000 e1000e igb ixgbe ixgbevf i40e vmxnet3 mlx5"
 # enabled drivers (bitfield)
 driver=
 drv()
--- a/LINUX/final-patches/mellanox--mlx5--3.3
+++ b/LINUX/final-patches/mellanox--mlx5--3.3
@@ -1,0 +1,379 @@
+diff --git a/mlx5/drivers/net/ethernet/mellanox/mlx5/core/Makefile b/mlx5/drivers/net/ethernet/mellanox/mlx5/core/Makefile
+index b8ce0b5..574d21a 100644
+--- a/mlx5/drivers/net/ethernet/mellanox/mlx5/core/Makefile
++++ b/mlx5/drivers/net/ethernet/mellanox/mlx5/core/Makefile
+@@ -1,12 +1,13 @@
+ ccflags-y += $(MLNX_CFLAGS)
+ 
+-obj-$(CONFIG_MLX5_CORE)		+= mlx5_core.o
++obj-$(CONFIG_MLX5_CORE)		+= mlx5_core$(NETMAP_DRIVER_SUFFIX).o
+ 
+-mlx5_core-y :=	main.o cmd.o debugfs.o fw.o eq.o uar.o pagealloc.o \
++mlx5_core$(NETMAP_DRIVER_SUFFIX)-y :=	main.o cmd.o debugfs.o fw.o eq.o uar.o pagealloc.o \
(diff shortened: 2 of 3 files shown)


giuseppelettieri:
Hi,

I have pushed the new branch, mlx5-bbc, with a tentative integration of Mellanox build system with our own. May you please give it a try? You should be able to compile with

```
./configure --drivers=mlx5
make
```

By the way, the driver fails to compile on 4.13 an beyond.

andrewbonney:
Thanks Giuseppe. I've tested that and can confirm it compiles and installs correctly. The driver brings the NIC online and packet transmission / reception appear to be working as expected.

I did also try adding a driver suffix but that didn't appear to work. Adding 'NETMAP_DRIVER_SUFFIX=@DRVSUFFIX@' to the build line as-per the Intel driver case seems to sort this out.

The driver we worked against was originally released in September 2016 so I'm not particularly surprised there's the odd issue against kernel 4.13. Hopefully once we've got this integrated it can be ported forwards onto some newer Mellanox driver versions, either by ourselves or other contributors.

giuseppelettieri:
Yep, I forgot about the suffix. Added now, and merged. Thanks!

Fr3DBr:
Hi @andrewbonney how are you ? May you let me know, if the ConnectX 3 Pro line, will work with netmap as well ?

kierank:
The ConnectX3 driver is different to the 4 and 5 and isn't supported by
netmap.

Kieran

PR #478, opened 2018-04-23
dkms: fix autoinstall

funman:
kernelver would not be set when calling dkms autoinstall --kernelver 4.13.0-39-generic,
and the build would use `uname -r`

This would break at kernel updates


funman:
ping

funman:
No problem :)

Btw we have started a irc channel : #netmap on freenode, that can be nice for real-time chat.

Feel free to join us!

PR #485, opened 2018-05-11
Mlx5: enable netmap mode

funman:
I noticed mlx5 driver is not being built with -DCONFIG_NETMAP

This pull request fixes this and also fixes build errors in the header.

Using pkt-gen locked up my machine in a couple seconds though.

@andrewbonney any comment on these build failures ?

I am guessing when you tested the build system patch from @giuseppelettieri the driver was actually in emulated mode rather than native.


andrewbonney:
I'll take a look on a fresh machine here and post back. Perfectly possible I've missed something in testing.

funman:
I removed the nm_config completely rather than use the ixgbe version.

Driver does not seem to lock the machine anymore

giuseppelettieri:
Thank you @funman, your patch is obviously correct. I'll delay the merge, however, until the other issue you have found has been figured out.

funman:
Ok no rush.
We'll also take a look at using a newer driver

andrewbonney:
Just to follow up here, apologies for the errors on my part. Our local version of netmap was dated 3rd January and I'd clearly forgotten to update it to the latest before submitting the initial PR. The changes here work for me having built locally, and there's a commit here which should resolve the nm_config issues matching up with more recent netmap modifications: https://github.com/bbc/netmap/commit/6ce986ad300a28413e3fe5df3244b6f11b34f716

Hopefully once @funman can also confirm things are working correctly this can all be merged in.

funman:
I am still not able to use the netmap driver, machine locks up really fast.

Need to gather more data and maybe test older versions

funman:
@andrewbonney this still works on ConnectX-4, right?

andrewbonney:
Yes, seems to work fine here with those patches against latest master

funman:
@vmaffione could you test with your ConnectX-5 ?

TX seems to work for me after I bump the # of descriptors to 4k, but after ^C I'll still need to force reset the computer.

funman:
Managed to get TX working with https://github.com/funman/netmap/commit/1bb933a34a99be57c6dd910963087aad185cc2bb

giuseppelettieri:
Let me know if you want to merge it now. I will not be able to test it by myself anytime soon.

funman:
I think you can merge this one then, I'll make another pull request with the other patches (they still need to be tested on ConnectX-4 though)

giuseppelettieri:
OK, thanks!

PR #486, opened 2018-05-11
Add checksum offload capability for emulated driver

andrewbonney:
The following patch adds support for checksum offloading to hardware when using the emulated driver function of netmap. In our use case where netmap originates traffic (rather than the main use case of forwarding) it's useful to have this mode available as a fallback.

There's a sysctl parameter added to enable this function, which remains disabled by default.

Assuming this looks generally acceptable as a contribution, I'd appreciate a review, not least in the following two areas:
* The first commit changes where the mbuf protocol is identified from (using the location of the ethertype rather than assuming IPv4)
* The second commit disables the offloading check in netmap.c when the new sysctl parameter is enabled. This is necessary to allow CHECKSUM_PARTIAL cases to pass, but also removed the check for GSO. It may be preferred to split the function nm_os_mbuf_has_offld into two so that the GSO check remains in place.

Thanks

### DIFF
--- a/LINUX/netmap_linux.c
+++ b/LINUX/netmap_linux.c
@@ -33,6 +33,7 @@
 #include <net/ip6_checksum.h>
 #include <linux/rtnetlink.h>
 #include <linux/nsproxy.h>
+#include <linux/ip.h>
 #include <net/pkt_sched.h>
 #include <net/sch_generic.h>
 #include <net/sock.h>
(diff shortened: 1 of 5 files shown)


vmaffione:
Looks good, but the use of `ETH_P_802_3_MIN`
causes compilation failures for many kernel versions: details here https://travis-ci.org/netmap-unipi/netmap/builds/379027397
Do we reallly need to validate the ethertype?

And yes, `nm_os_mbuf_has_offld` could be splitted as you suggest.

andrewbonney:
Thanks. I've split the offload check function - let me know if the original function signature needs maintaining for any reason.

For the ethertype, it looks like in order to set `m->protocol` correctly there's a built in function `eth_type_trans` which should be available on all kernel versions, but it is intended for packet reception so requires a call to fix the mbuf data pointer and length afterwards. I've modified the patch to use this for now.

For the checksum additions we only really need to update the setting of `m->protocol` to detect both IPv4 and IPv6, but having encountered issues with some NIC drivers in the past it seemed to make sense to set this value as accurately as possible. Let me know what you think.

vmaffione:
I wouldn't call `eth_type_trans`. It's a receive function, it sets skb fields that we shouldn't set, and it is also expensive. Also, it forces use to call `skb_push`, and this is cumbersome.

If it's important to make sure that `ethertype >= ETH_P_802_3_MIN`, then we can just define the macro ourselves (`#define NM_ETH_P_802_3_MIN 0x0600`), and use that. In this way we can use the previous version of the code.

vmaffione:
Thanks. Travis is also happy with the change.
Merging in a little while.

PR #520, opened 2018-07-31
Fix FreeBSD PR206053 - panic when adding netmap device to kqueue.

dabright:
See FreeBSD PR206053 (https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=206053) for details.

### DIFF
--- a/sys/dev/netmap/netmap_freebsd.c
+++ b/sys/dev/netmap/netmap_freebsd.c
@@ -1502,7 +1502,7 @@ netmap_kqfilter(struct cdev *dev, struct knote *kn)
 	kn->kn_fop = (ev == EVFILT_WRITE) ?
 		&netmap_wfiltops : &netmap_rfiltops;
 	kn->kn_hook = priv;
-	knlist_add(&si->si.si_note, kn, 1);
+	knlist_add(&si->si.si_note, kn, 0);
 	// XXX unlock(priv)
 	ND("register %p %s td %p priv %p kn %p np_nifp %p kn_fp/fpop %s",
 		na, na->ifp->if_xname, curthread, priv, kn,


dabright:
The problem this fixes is now also logged as netmap issue #521.

vmaffione:
Thanks a lot! What kind of testing did you do to check that it is now fixed? Do you have any test program you can share?

dabright:
I used a slight modification of the test program submitted as a comment to the FreeBSD PR. Before the change, running the program (on -CURRENT) produced a kernel panic essentially identical to that shown in the PR. After the change, the program runs to completion without panic. Note that I am not sure how to force the traffic on the netmap device for which the kqueue is waiting in the program, so it simply timed out. However, I was mainly interested in just fixing the panic.

Hmmm, it seems I can't attach the program file, so I'll just list it inline below:
```c
#include <stdio.h>
#include <assert.h>

#include <sys/types.h>
#include <sys/event.h>
#include <sys/time.h>

#ifndef NETMAP_WITH_LIBS
#define NETMAP_WITH_LIBS
#endif
#include <net/netmap_user.h>

static int
kq_add(int kq, int fd)
{
	struct kevent changes[1];
	int ret;

	EV_SET(&changes[0], fd, EVFILT_READ, EV_ADD, 0, 0, NULL);
	ret = kevent(kq, changes, 1, NULL, 0, NULL);
	assert(ret != -1);

	printf("[%s] success\n", __func__);

	return (ret);
}

static void
kq_wait(int kq)
{
	struct kevent events[1];
	int ret;
	struct timespec to = {1, 0};

	ret = kevent(kq, NULL, 0, events, 1, &to);
	assert(ret != -1);

	printf("[%s] success\n", __func__);
}

int main(void)
{
	const char *ifname = "vale0:vm1";
	struct nm_desc *d;
	int kq;

	d = nm_open(ifname, NULL, 0, 0);
	assert(d != NULL);

	kq = kqueue();
	assert(kq != -1);

	kq_add(kq, d->fd);
	kq_wait(kq);

	return (0);
}
```

vmaffione:
I see, thanks. Yes, your program is not reading from the netmap ring and nobody is sending anything to the `vale0:vm1` port. So a timeout is expected.
We can easily write a test program to test kqueue, starting from this example.

Issue #433, opened 2018-01-15
virtio: support single-queue mode


carlgsmith:
In our application we would like to be able to independently open the RX and TX rings as well as the physical and host rings of an interface using the native virtio driver.
i.e. eth0/R, eth0^/T, eth0^/R, eth0/T

Actually the real use case to only monitor ingress, eth0/R -> eth0^/T, and completely ignore egress.

The comment in the code is:
	/* These virtio-net driver patches do not support single-queue mode
	 * (modifications would be needed to free_unused_bufs()
	 * free_receive_bufs()). As a result, we fail here if we detect
	 * the user is trying to open or close only a subset of the rings. */

Has anyone attempted to address this? By chance has someone got some patches lying around that might be able to do it? Or some pointers on where to start .....

vmaffione:
Hi,
  This limitation can be easily removed if you stick to a particular version of the virtio_net.c driver.

You'd need to provide a version of `free_unused_bufs` (http://elixir.free-electrons.com/linux/v4.0/source/drivers/net/virtio_net.c#L1478) that processes only a single TX queue or a single RX queue. Same thing for `free_receive_bufs`, `virtio_netmap_clean_used_ring` and `virtio_netmap_reclaim_unused`, and `virtio_netmap_init_sg`. Once you have all these functions, you can call them in `virtio_netmap_reg` to enter/exit netmap mode only on specific TX and/or RX queues.

The reason why this has not been done is that `free_unused_bufs` and `free_receive_bufs` change wildly across Linux versions, so we would need to maintain tens of versions for 6-7 functions, which means a lot of code. And the current virtio-net patch is already very complex to read and mantain (again, because of implementation details that vary across code).

An alternative approach to avoid all of this fuss would be something similar to what we do for e1000e, ixgbe, ixgbevf, igb and i40e drivers: we stick to a version of virtio_net.c (e.g. 4.14) and make a netmap patch only for that version, so that we can use the same virtio-net driver also for previous kernel versions. Of course a 4.14 driver would not compile down to 2.6.32, but maybe it covers at least all the 4.x versions.

carlgsmith:
As a quick test we commented out the early return in virtio_netmap_reg to see how far things got. Surprisingly it all seems to work for our case... (separately opening only eth0/R and eth0^/T).

We are able to pick up packets from eth0/R and forward them on to eth0^/T.

When packets are transmitted by the host on eth0/T, this code int netmap_transmit is hit;
[...]

PR #527, opened 2018-08-07
Virtio All TX or All RX

sdpetrides:
Based on issue [443](https://github.com/luigirizzo/netmap/issues/433)

I wanted to be able to use all TX ring or all RX rings for virtio. I was not sure how to tell which rings were "pending" from the `virtnet_info` struct, so I would pass in TX, RX, or TXRX as an additional flag. There may be a better way to do this.

General changes:

- `virtio_netmap_reg` in `virtio_netmap.h` now initializes TX and RX separately. Subroutines have flag (`enum txrx`) that indicates TX, RX or TXRX.
- Patch (`vanilla-virtio-40f00-41000`) for `virtio_net.c` is updated to use the same TX/RX init separation flag

Configure and make commands:
`./configure --kernel-sources=/usr/src/kernels/linux-stable-4.15 --drivers=virtio_net.c`
`make`

`linux-stable-4.15` is taken from [kernel.org](https://www.kernel.org)

Results:

I have tested with TX only and I am able to receive traffic through the kernel and send packets through netmap-virtio. However, after large amounts of traffic (100MB, 1G, etc.), the port no longer receives packets. I have not found the reason for this. As of now, this is this is a proof-of-concept. Looking for feedback 😄

Edit: fix link


carlgsmith:
Have pushed our version of these changes to
https://github.com/alliedtelesis/netmap/commits/virtio_single_queue

We have not rebased this cycle, so they may be a touch out of date.

Note the top patch was the one we need to avoid the "getting stuck after a while" problem. It was caused by the queues being reconfigured when the second open was made.

sdpetrides:
Hey Carl,

Thanks for pushing your code. I have found we have some similarities in our patches. I have tried to build and test with your patch but I am having trouble during configuration. I think we have different kernel versions. With `4.15.18` I am getting the following issue:

```
*** There were problems patching the driver virtio_net.c.
***
*** Disabling virtio_net.c.
```

Can you post more info on your build parameters and kernel sources?

sdpetrides:
I have played around with this more but I am still having an issue after enough traffic has been sent. Specifically, after enough traffic has been sent, the kernel stack does not read anything (from sockets). My application behaves as if there is no incoming data. However, if I exit netmap mode and use the linux stack normally (send and recv with sockets) and then switch back to netmap mode in a separate process, the issue is gone and I can read from sockets and send through netmap.

I am curious to learn more about your issue with "other event generating a softirq" to see if this applied to my issue. Again, I appreciate your help.

carlgsmith:
I think we did have that issue and we had to modify our application to avoid it. Check your ring pointers when it is stalled.

One change we made was to ensure we never pushed any packets to the host TX ring unless there was at least 2 spaces;

e.g.
  /* Lock ring and wait for room if needed */
  try_again:
    pthread_mutex_lock (&ring->lock);
    txring = NETMAP_TXRING (ring->desc->nifp, ring_id);
    if (nm_ring_space (txring) < 2)
    {
        VERBOSE ("PKT(WAIT): oif:%s[%u] TX Ring Full\n", oif->ifname, ring_id);
        ioctl (ring->desc->fd, NIOCTXSYNC, 0);
        usleep (1);
        pthread_mutex_unlock (&ring->lock);
        goto try_again;
    }

We found if we did not do this, it blocked up completely.

The other common problem is if you do not follow the ownership model properly (e.g. multiple threads), you can get the ring pointers into funny states.

sdpetrides:
I think we may be having a similar but different issue since we do not pass any packets through to the TX rings of the host stack. Your tip to look at the ring pointers is good and I am looking into that now.

sdpetrides:
I have dug into the issue a bit more and found that there my issues during "large file transfers" are not an issue with "large file transfers" but an issue with ARP requests that trigger a `netmap_transmit()`. I do not handle these and the traffic must be dropped or sent to the wrong place. I will have to find a way to handle this in my app.

Thank you again for your help.

PR #533, opened 2018-08-15
Memset fix

stephen-hurd:
Fixes a harmless transposition of memset() parameters.  To be honest, I don't think the memset() is accomplishing anything here anyway.


vmaffione:
Thanks!

PR #537, opened 2018-09-02
remove docs about how kqueue is unsupported

jmgurney:
Looks like old docs says that kqueue is unsupported.  Multiple other places in the page says that it is supported.


vmaffione:
Yes, kqueue and epoll are supported. Thanks.

PR #547, opened 2018-09-26
netmap/netmap_mem2: fix memory pool bitmap not initialized all bits

DongCai008:
memset is based on number of bytes not the sizeof(u32). therefore, not all bitmap bytes initialized


vmaffione:
Thanks.

PR #558, opened 2018-11-08
NAF_FORCE_RECLAIM can cause no ring buffers to be available

joshraiff-sophos:
When doing a force reclaim, the code blindly moves nr_hwtail.
It needs to ensure there is at least one free buffer.

It looks like all the drivers have this change.  I can do that if we decide this is the correct change.

### DIFF
--- a/LINUX/ixgbe_netmap_linux.h
+++ b/LINUX/ixgbe_netmap_linux.h
@@ -421,10 +421,13 @@ ixgbe_netmap_txsync(struct netmap_kring *kring, int flags)
 #ifndef NM_IXGBE_USE_TDH
 	(void)reclaim_tx;
 	if ((flags & NAF_FORCE_RECLAIM) || nm_kr_txempty(kring)) {
+		u_int tail_i;
 		nic_i = NM_ACCESS_ONCE(*ina->heads[ring_nr].phead);
 		nm_i = netmap_idx_n2k(kring, nic_i);
 		ND(5, "%s: h %d", kring->name, h);
-		kring->nr_hwtail = nm_prev(nm_i, lim);
+		tail_i = nm_prev(nm_i, lim);
+		if (tail_i != kring->nr_hwcur)
+			kring->nr_hwtail = tail_i;
 	}
 #else /* NM_IXGBE_USE_TDH */
 	/*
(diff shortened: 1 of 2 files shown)


giuseppelettieri:
The code was updating the ```nr_hwtail``` pointer based on the value coming from the NIC. Have you actually observed any error caused by this?

joshraiff-sophos:
Yes.  I have a setup where I have a vale switch that routing traffic to the host and to a user app.  After bringing it up, if I ping the ip address associated with the host port, I can see an arp request coming in, the host generates an arp reply, but it never goes out. 

I added some debug prints and found that the curr and tail pointers were the same after returning from ixgbe_netmap_txsync() so nm_kr_space() is returning 0 buffers available.  The igb driver does not exhibit this behavior.

joshraiff-sophos:
Perhaps I am treating a symptom of another issue?  I've boiled reproducing this to a fairly small set of commands.  If I revert https://github.com/luigirizzo/netmap/commit/303763f99457532dd089faabf72dabe2bdcad811 I do not see this bug.

insmod ./netmap.ko
insmod ./ixgbe/ixgbe.ko
ethtool -K Port9 tx off gro off gso off tso off sg off rxvlan off txvlan off
vale-ctl -h vale0:Port9
ifconfig Port9 up
vale-ctl -n host1
vale-ctl -a vale0:host1 -m 1
sysctl -w net.ipv4.ip_forward=1
net.ipv4.ip_forward = 1
ip link set dev Port9 up
ip addr add 1.1.1.1/16 dev Port9
ping 1.1.1.2 -c 1
PING 1.1.1.2 (1.1.1.2) 56(84) bytes of data.
64 bytes from 1.1.1.2: icmp_seq=1 ttl=64 time=2027 ms

--- 1.1.1.2 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 2027.677/2027.677/2027.677/0.000 ms
ip link set dev Port9 down   ####  Setting the link down triggers the issue.
ip link set dev Port9 up
ping 1.1.1.2 -c 1
PING 1.1.1.2 (1.1.1.2) 56(84) bytes of data.
From 1.1.1.1 icmp_seq=1 Destination Host Unreachable

--- 1.1.1.2 ping statistics ---
1 packets transmitted, 0 received, +1 errors, 100% packet loss, time 0ms

PR #569, opened 2018-11-27
Upstream vale port type extraction

garrettkajmowicz-sophos:
Adds some logic which we have found to be useful. Namely, provide a way to determine what type of VALE port we are dealing with. This is useful to identify configuration issues or to be able to take different actions based on where the data is coming from.


sys/dev/netmap/netmap_legacy.c
```diff
@@ -324,6 +324,7 @@ nmreq_to_legacy(struct nmreq_header *hdr, struct nmreq *nmr)
 		strlcpy(nmr->nr_name, hdr->nr_name, sizeof(nmr->nr_name));
 		nmr->nr_arg1 = req->nr_bridge_idx;
 		nmr->nr_arg2 = req->nr_port_idx;
+		nmr->nr_arg3 = req->nr_port_type;
```
vmaffione: This is legacy API, and therefore can't be changed.

vmaffione:
Hi,
  We discussed with @giuseppelettieri , and we think it is necessary to change the approach to play with this kind of port-type information.
Trying to recover the port type by peeking inside pointers of the netmap adapters is very fragile (w.r.t. changes) and unnecessarily complex. Also, we should make this mechanism more generic, and make it possible to recover the port-type also for non-VALE ports.

A better approach is to add an `adapter_type` or `port_type` enum field to the `struct netmap_adapter`, so that any netmap port has a port type. The `adapter_type` field is set when the netmap adapter struct is created. This may be `netmap_attach` for NIC ports, `netmap_get_null_na` for null ports, `netmap_get_pipe_na` for pipe ports, etc. You can look at the `netmap_get_na` function to see all the ports that we support.

Once we do this, we can extend the `struct nmreq_port_info_get` to retrieve the additional `adaper_type` field. We can also add this field to `struct nmreq_vale_list`, as your original change did.

PR #570, opened 2018-11-30
Mellanox 4.5 ConnectX-5

funman:
This PR adapts the patch for Mellanox 3.3 tested on ConnectX-4, to the 4.5 driver tested with ConnectX-5.

I would like to get some feedback before I do this and squash the commits together.
I left a couple of TODOs, especially in places where we have to check if a queue is using netmap or not.

Note: it is not possible to select 3.3 or 4.5, because the source tarballs have different format (in both cases it's a tgz inside a tgz inside a tgz ..........)

Comments welcome!


funman:
I updated this PR with a patch for recently released 4.5 driver, I pushed the previous branch content in my mlx5-drv-4.3 branch

giuseppelettieri:
I haven't found the time to test this yet, but I see that others have done it. Many thanks!

PR #574, opened 2018-12-07
Add ability to change MTU once device driver is loaded.

joshraiff-sophos:
If the linux interface device driver supports NS_MOREFRAG or the
new MTU still fits within a netmap buffer, allow the MTU to be
changed.

### DIFF
--- a/sys/dev/netmap/netmap.c
+++ b/sys/dev/netmap/netmap.c
@@ -2117,6 +2117,53 @@ netmap_csb_validate(struct netmap_priv_d *priv, struct nmreq_opt_csb *csbo)
 	return 0;
 }
 
+/* Ensure that the netmap adapter can support the given MTU.
+ * @return EINVAL if the na cannot be set to mtu, 0 otherwise.
+ */
+int
+netmap_buf_size_validate(const struct netmap_adapter *na, unsigned mtu) {
+	unsigned nbs = NETMAP_BUF_SIZE(na);
+
+	if (mtu <= na->rx_buf_maxsize) {
+		/* The MTU fits a single NIC slot. We only
+		 * Need to check that netmap buffers are
+		 * large enough to hold an MTU. NS_MOREFRAG
+		 * cannot be used in this case. */
+		if (nbs < mtu) {
+			nm_prerr("error: netmap buf size (%u) "
+				 "< device MTU (%u)", nbs, mtu);
+			return EINVAL;
+		}
+	} else {
+		/* More NIC slots may be needed to receive
+		 * or transmit a single packet. Check that
+		 * the adapter supports NS_MOREFRAG and that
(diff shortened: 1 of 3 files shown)


giuseppelettieri:
I think we can merge this. What do you think, @vmaffione?

vmaffione:
The point of current code is to prevent the MTU from changing while the NIC is in netmap mode. If this happens, netmap applications already in execution may try to transmit packets larger than the MTU.
We do not want to allow that, I think.

You can actually change the MTU while the NIC is not in netmap mode. Why is this not sufficient?

joshraiff-sophos:
I have all of my devices attached to a vale switch, its not so easy to move out of netmap mode without detaching them is it?  What happens to in-flight on a non-netmap NIC when you change the MTU?

vmaffione:
Yes, in that use-case it may be useful.
MTU changes while a netdevice is up is supported by Linux and FreeBSD network stacks.
I'm trying to think about what could be possibly go wrong if you change the MTU while netmap applications are running, but I don't really see a problem (since you validate).

sys/dev/netmap/netmap.c
```diff
@@ -2117,6 +2117,54 @@ netmap_csb_validate(struct netmap_priv_d *priv, struct nmreq_opt_csb *csbo)
 	return 0;
 }
 
+/* Validate the na can accomodate the given MTU.
+ * @return EINVAL if the NA cannot be set to mtu
+ *         0 if it can.
+ */
+int
+netmap_mtu_validate(const struct netmap_adapter *na, unsigned mtu) {
```
vmaffione: This is more `netmap_buf_size_validate`, because that's what we are validating (MTU is given).

sys/dev/netmap/netmap.c
```diff
@@ -2117,6 +2117,54 @@ netmap_csb_validate(struct netmap_priv_d *priv, struct nmreq_opt_csb *csbo)
 	return 0;
 }
 
+/* Validate the na can accomodate the given MTU.
```
vmaffione: Rephrase:
````
Ensure that the netmap adapter can support the given MTU.
````

File: sys/dev/netmap/netmap.c [hunk 3]
```diff
@@ -2117,6 +2117,54 @@ netmap_csb_validate(struct netmap_priv_d *priv, struct nmreq_opt_csb *csbo)
...
+ * @return EINVAL if the NA cannot be set to mtu
```
vmaffione:
```suggestion
 * @return EINVAL if the na is not compatible with the given mtu, 0 otherwise.
```

vmaffione:
Can you please squash the two commits?

vmaffione:
yes, you can force push to your branch, and rebase if you wish

PR #613, opened 2019-04-30
Make sure num_rx is correctly initialized

j-t-d:
nmport_mmap doesn't correctly initialize num_rx before using it to determine first_rx_ring


giuseppelettieri:
Thanks for catching this.

PR #614, opened 2019-05-05
bdg config interface locking

YutaroHayakawa:
Since bdg config interface is supposed to be used for updating the configuration of the module (e.g. updating the routing table), I think it should take writer lock instead of reader lock.


YutaroHayakawa:
Seems this is outdated. Let me close.

Issue #602, opened 2019-03-21
Segmentation fault when trying to allocate 2 million extra buffers


mkaniewski:
Hi,

I have a program which allocates a lot of extra netmap buffers. When there is a lot of them requested ( > 2 millions) the netmap returns them succesfully but I get a segmentation fault when I try to dereference the buffer pointer. I prepared a sample application that shows the problem:

```
#include <sys/types.h>
#include <sys/event.h>
#include <sys/time.h>

#include <assert.h>
#include <stdlib.h>
#include <stdio.h>

#define NETMAP_WITH_LIBS
#pragma GCC diagnostic ignored "-Wcast-qual"
#include <net/netmap_user.h>

static void
usage()
{
        fprintf(stderr, "Usage: niocregif -i iface.\n");
        exit(1);
}

int main(int argc, char *argv[])
{
        struct nmreq req;
        struct netmap_if *nifp;
        struct netmap_ring *ring;
        size_t currindex;
        int ch, error, fd;
        const char *ifacestr;
        void *p;

        ifacestr = NULL;
        while ((ch = getopt(argc, argv, "i:")) != -1) {
                switch (ch) {
                case 'i':
                        ifacestr = optarg;
                        break;
                default:
                        usage();
                }
        }

        if (ifacestr == NULL) {
                usage();
        }

        fd = open("/dev/netmap", O_RDWR);
        if (fd == -1) {
                D("Unable to open /dev/netmap");
                return -1;
        }

        bzero(&req, sizeof(req));

        memcpy(req.nr_name, ifacestr, strlen(ifacestr));
        req.nr_version = NETMAP_API;
        req.nr_ringid = NETMAP_NO_TX_POLL | 0;
        req.nr_flags = NR_REG_ONE_NIC;
        req.nr_arg3 = 2000000;

        printf("Opening netmap device %s.\n", ifacestr);
        error = ioctl(fd, NIOCREGIF, &req);
        assert(error == 0);

        p = mmap(0, req.nr_memsize, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
        assert(mmap != MAP_FAILED);
        nifp = NETMAP_IF(p, req.nr_offset);
        assert(nifp != NULL);
        ring = NETMAP_RXRING(nifp, 0);

        currindex = nifp->ni_bufs_head;
        for (unsigned int ii = 0; ii < req.nr_arg3; ii++) {
                char *buf;
                size_t *nextindex;

[...]

PR #637, opened 2019-07-30
Use size_t instead of u_int for total memory.

j-t-d:
In testing, when using buf_num over 420,000 or so, with the default
clusternumbers, we run into integer overflow and segfault. 

In our environment, we have 40 core servers (20+20) with the nics using
40 tx/rx rings, with the ring slot size set to 4096. We have to set the
buf_num to about 750000 in order to allocate enough memory for our
application to load with around 100000 extra buffers.

This segfaults without this change.

This is basically adapted from issue #602. I left the max buffers at the
default of 1_000_000 and changed some %lu and %ld to %zu %zd

### DIFF
--- a/sys/dev/netmap/netmap_mem2.c
+++ b/sys/dev/netmap/netmap_mem2.c
@@ -100,16 +100,17 @@ struct netmap_obj_pool {
 	/* ---------------------------------------------------*/
 	/* these are only meaningful if the pool is finalized */
 	/* (see 'finalized' field in netmap_mem_d)            */
-	u_int objtotal;         /* actual total number of objects. */
-	u_int memtotal;		/* actual total memory space */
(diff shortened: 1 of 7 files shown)


sys/dev/netmap/netmap_mem2.c
```diff
@@ -100,16 +100,17 @@ struct netmap_obj_pool {
 	/* ---------------------------------------------------*/
 	/* these are only meaningful if the pool is finalized */
 	/* (see 'finalized' field in netmap_mem_d)            */
-	u_int objtotal;         /* actual total number of objects. */
-	u_int memtotal;		/* actual total memory space */
```
j-t-d: Just a note here, I reorganized the struct a bit to remove a 4 byte hole.

PR #651, opened 2019-09-25
remove compiler warnings when building for arm32

ieio74:
Small modifications to build netmap.ko on arm32 without compiler warnings
Tested arm32 platform: imx6
Compiler: gcc-linaro-7.3.1-2018.05 (https://releases.linaro.org/components/toolchain/binaries/7.3-2018.05/arm-linux-gnueabihf/)

Steps I followed:
1. ./configure --no-drivers --no-ext-drivers --kernel-sources=$PATH2KERNELSRC --kernel-dir=$PATH2KERNELSRC --kernel-opts="ARCH=arm CROSS_COMPILE=/opt/gcc-linaro-7.3.1-2018.05-x86_64_arm-linux-gnueabihf/bin/arm-linux-gnueabihf-" --cc=/opt/gcc-linaro-7.3.1-2018.05-x86_64_arm-linux-gnueabihf/bin/arm-linux-gnueabihf-gcc --ld=/opt/gcc-linaro-7.3.1-2018.05-x86_64_arm-linux-gnueabihf/bin/arm-linux-gnueabihf-ld --driver-suffix=-netmap

2. make netmap.ko

### DIFF
--- a/LINUX/netmap_linux.c
+++ b/LINUX/netmap_linux.c
@@ -1824,8 +1824,11 @@ nm_os_pt_memdev_iomap(struct ptnetmap_memdev *ptn_dev, vm_paddr_t *nm_paddr,
 	*mem_size = ioread32(ptn_dev->pci_io + PTNET_MDEV_IO_MEMSIZE_HI);
 	*mem_size = ioread32(ptn_dev->pci_io + PTNET_MDEV_IO_MEMSIZE_LO) |
 		(*mem_size << 32);
-
+#ifdef __arm__
+	nm_prinf("=== BAR %d start %x len %x mem_size %lx ===",
+#else
 	nm_prinf("=== BAR %d start %llx len %llx mem_size %lx ===",
+#endif
 			PTNETMAP_MEM_PCI_BAR,
 			pci_resource_start(pdev, PTNETMAP_MEM_PCI_BAR),
 			pci_resource_len(pdev, PTNETMAP_MEM_PCI_BAR),
--- a/LINUX/netmap_ptnet.c
+++ b/LINUX/netmap_ptnet.c
@@ -1270,7 +1270,11 @@ ptnet_probe(struct pci_dev *pdev, const struct pci_device_id *ent)
 	}
 
 	err = -EIO;
+#ifdef __arm__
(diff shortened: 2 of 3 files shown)


LINUX/netmap_linux.c
```diff
@@ -1824,8 +1824,11 @@ nm_os_pt_memdev_iomap(struct ptnetmap_memdev *ptn_dev, vm_paddr_t *nm_paddr,
 	*mem_size = ioread32(ptn_dev->pci_io + PTNET_MDEV_IO_MEMSIZE_HI);
 	*mem_size = ioread32(ptn_dev->pci_io + PTNET_MDEV_IO_MEMSIZE_LO) |
 		(*mem_size << 32);
-
+#ifdef __arm__
+	nm_prinf("=== BAR %d start %x len %x mem_size %lx ===",
+#else
```
jhk098: Instead of this, would it work to explicitely cast the return value of pci_resource_start and pci_resource_len to (unsigned long long)?

LINUX/netmap_ptnet.c
```diff
@@ -1270,7 +1270,11 @@ ptnet_probe(struct pci_dev *pdev, const struct pci_device_id *ent)
 	}
 
 	err = -EIO;
+#ifdef __arm__
+	pr_info("%s: IO BAR (registers): start 0x%x, len %u, flags 0x%lx\n",
+#else
 	pr_info("%s: IO BAR (registers): start 0x%llx, len %llu, flags 0x%lx\n",
```
jhk098: Same as above

LINUX/netmap_linux.c
ieio74: Well, if I well understood this is used by ptnet that is related to QEMU/KVM. 
I currently don't have a setup to test it, but I think that casting the two returns to unsigned long long it will work as well.
As it removes compiler warnings as well, so up to me it is more than fine.

LINUX/netmap_ptnet.c
ieio74: also for the reply, same as above.

jhk098:
Ok, so if current master works we can close this pull request.

PR #661, opened 2019-11-04
Record mapped netmap memory as RssShmem rather than RssFile

carlgsmith:
VmRSS     size of memory portions. It contains the three
           following parts (VmRSS = RssAnon + RssFile + RssShmem)
 RssAnon   size of resident anonymous memory
 RssFile   size of resident file mappings
 RssShmem  size of resident shmem memory (includes SysV shm,
           mapping of tmpfs and shared anonymous mappings)

The kernel differentiates between RssFile and RssShmem by
checking the PG_swapbacked flag on the page. By setting this
flag the netmp memory will be recorded as RssShmem.


giuseppelettieri:
thanks!

PR #669, opened 2020-01-07
veth: fix segfault at access to not created adapter rings

kogdenko:
Remove peer in veth_netmap_krings_delete(vna) because without that
subsequent call of veth_netmap_krings_create(vna->peer) do not create rings.

### DIFF
--- a/LINUX/veth_netmap.h
+++ b/LINUX/veth_netmap.h
@@ -81,6 +81,7 @@ veth_netmap_dtor(struct netmap_adapter *na)
 		(struct netmap_veth_adapter *)na;
 	if (vna->peer_ref) {
 		vna->peer_ref = 0;
+		vna->peer->peer = NULL;
 		netmap_adapter_put(&vna->peer->up.up);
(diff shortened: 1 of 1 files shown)


LINUX/veth_netmap.h
```diff
@@ -81,6 +81,7 @@ veth_netmap_dtor(struct netmap_adapter *na)
 		(struct netmap_veth_adapter *)na;
 	if (vna->peer_ref) {
 		vna->peer_ref = 0;
+		vna->peer->peer = NULL;
```
giuseppelettieri: do we need to reset ```vna->peer``` too?

kogdenko: No - it is not necessary to set  vna->peer->peer for fixing bug, but I think it is better to remove reference to destructed peer.

giuseppelettieri: OK. I have reproduced the bug and confirmed that your patch fixes it. Thanks!

PR #700, opened 2020-05-22
fix NETMAP_WOFFSET and typos

guruofquality:
@giuseppelettieri I fixed the macro from https://github.com/luigirizzo/netmap/commit/a2b1b4ea221a63626ff4b1bfa07c6c356da28434#commitcomment-39365274

I also found a few typos in libnetmap.h, I hope that you dont mind

Note that I didn't actually test the offset support. I'm just experimenting with some of the netmap features. I'm using i40e and it looks like offset is just implemented in e1000


giuseppelettieri:
merged, thanks!

The offset feature should be available also in most other Intel drivers, unless something is not working properly.

PR #712, opened 2020-07-31
fix parsing of legacy nmr->nr_ringid

brian90013:
Code was checking for NETMAP_{SW,HW}_RING in req->nr_ringid which
had already been masked by NETMAP_RING_MASK. Therefore, the comparisons
always failed and set NR_REG_ALL_NIC. Check against the original nmr
structure.


giuseppelettieri:
good catch, thanks!

PR #721, opened 2020-09-09
linux/nm_os_vi_persist: Create new interface in net_ns of current task

### LINKED ISSUE #720: Issues running lb in non-root network namespace on Linux
Hello,

I'm experimenting with `netmap` and am attempting to run a "pipe-only" `lb` inside a new network namespace:
```
$ sudo unshare -n
$ lb -i in}0 -p out:4
171.051996 main [671] interface is in}0
171.054191 nm_open [965] NIOCREGIF failed: No such device or address netmap:in}0
171.054199 main [802] cannot open netmap:in}0
```

The `in` device does actually appear in the root network namespace (where it can't be deleted, but starting and stopping lb in the root network namespace deletes it):
```
# In root network namespace
$ ip link show in
36: in: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1
    link/ether 86:2b:52:c2:5b:c4 brd ff:ff:ff:ff:ff:ff
$ sudo ip link del in
RTNETLINK answers: Operation not supported

# This cleans up the `in` interface, phew..
$ lb -i in}0 -p out:4
ctrl+c
```

Should `lb` generally be expected to work in a non-root network namespace on Linux when used as above?

I've come up with the following patch and this seems to solve the isolated `lb` case, but I've fairly little knowledge if this is the right thing to do (let me know if you'd rather have a PR for this): 
```
diff --git a/LINUX/netmap_linux.c b/LINUX/netmap_linux.c
index 8a4ea5b1..2dfe6138 100644
--- a/LINUX/netmap_linux.c
+++ b/LINUX/netmap_linux.c
@@ -2358,7 +2358,7 @@ nm_os_vi_persist(const char *name, struct ifnet **ret)
                error = ENOMEM;
                goto err_put;
        }
-       dev_net_set(ifp, &init_net);
+       dev_net_set(ifp, current->nsproxy ? current->nsproxy->net_ns : &init_net);
        ifp->features |= NETIF_F_NETNS_LOCAL; /* just for safety */
        ifp->dev.driver = &linux_dummy_drv;
        error = register_netdev(ifp);
```

awelzel:
May fix #720 .

### DIFF
--- a/LINUX/netmap_linux.c
+++ b/LINUX/netmap_linux.c
@@ -2358,8 +2358,10 @@ nm_os_vi_persist(const char *name, struct ifnet **ret)
 		error = ENOMEM;
 		goto err_put;
 	}
-	dev_net_set(ifp, &init_net);
+#ifdef CONFIG_NET_NS
+	dev_net_set(ifp, current->nsproxy->net_ns);
 	ifp->features |= NETIF_F_NETNS_LOCAL; /* just for safety */
+#endif
 	ifp->dev.driver = &linux_dummy_drv;
 	error = register_netdev(ifp);
 	if (error < 0) {


giuseppelettieri:
Thanks!

PR #725, opened 2020-09-12
Add mlx5 version 5.1 patch

kierank:


LINUX/configure
```diff
@@ -504,7 +504,7 @@ SRCDIR=$SRCDIR
 ifneq (\$(KERNELRELEASE),)
 obj-m := $TESTOBJS
 else
-EXTRA_CFLAGS := -Wno-unused-variable -Wno-unused-label -Werror \$(addprefix -Wno-error=,$REC_DISABLED_WARNINGS)
+EXTRA_CFLAGS := -Wframe-larger-than=2000 -Wno-maybe-uninitialized -Wno-unused-variable -Wno-unused-label -Werror \$(addprefix -Wno-error=,$REC_DISABLED_WARNINGS)
```
kierank, line 507: I would like to know if there is a better way of doing this to work around warnings in the mlx5 driver

giuseppelettieri: You should add the `no-` ones to the `DISABLED_WARNINGS` variable in `LINUX/configure`. The script will filter-out the warnings that are not recognized by the installed compiler and put the remaining ones into `REC_DISABLED_WARNINGS`. This will not work for `-Wframe-larger-than=2000`, so you should leave it as-is in `EXTRA_CFLAGS`.

kierank:
Thanks, there is now an mlx5 version 5.2 out so I will send a patch for that soon

giuseppelettieri:
Looking forward to it!

PR #728, opened 2020-09-27
Fix constness warnings generated when "-Wcast-qual" compiler option is used.

mkaniewski:
Compiler generates some warnings when `-Wcast-qual` option is used and `netmap_user.h` is included in a program. This pull request fixes the problematic code.
It also fixes a simple error in the example code presented in `netmap(4)`.


giuseppelettieri:
Thanks!

PR #731, opened 2020-10-02
octeon: Add Cavium Octeon driver support

carlgsmith:


LINUX/octeon_netmap.h
```diff
@@ -0,0 +1,498 @@
...
+	/* Enable or disable */
+	if (onoff) {
+		if (na->active_fds == 0) {
+			octeon_netmap_enable_port(na);
+		}
+		for_rx_tx(t) {
+			for (i = 0; i <= nma_get_nrings(na, t); i++) {
+				struct netmap_kring *kring = NMR(na, t)[i];
+				if (nm_kring_pending_on(kring)) {
+					kring->nr_mode = NKR_NETMAP_ON;
```
jhk098, line 243: For hardware NICs, this operation is usually performed by calling `netmap_reset` from the driver patch. You can look at intel drivers, for example, or vmxnet3.

LINUX/octeon_netmap.h
```diff
@@ -0,0 +1,498 @@
...
+			}
+		}
+		nm_set_native_flags(na);
+	} else {
+		nm_clear_native_flags(na);
+		for_rx_tx(t) {
+			for (i = 0; i <= nma_get_nrings(na, t); i++) {
+				struct netmap_kring *kring = NMR(na, t)[i];
+				if (nm_kring_pending_off(kring)) {
+					kring->nr_mode = NKR_NETMAP_OFF;
```
jhk098, line 254: Same as above.

LINUX/octeon_netmap.h
carlgsmith: Done

LINUX/octeon_netmap.h
carlgsmith: Done

carlgsmith:
Have also updated patch to include NAF_OFFSETS support for Octeon

giuseppelettieri:
Hi, sorry for the late reply. We are more than willing to merge this, but vanilla patches are auto-generated from a different repository which is a clone of vanilla linux. You find a copy of the repo here: [git@github.com:giuseppelettieri/linux-pspat.git](url). There is a ```netmap-$v``` branch for each version of linux. You should remove the ```vanilla--octeon--*``` file from this pull request and create a pull request for the other repo, containing only the necessary changes for the driver. You can target any one of the ```netmap-$v``` branches. I will then try to forward/backword port the commit as much as possible and then auto-generate the patch for the netmap repository.

PR #733, opened 2020-10-22
freebsd: fix mutex double unlock

brian90013:
Hello,

I am testing FreeBSD-12.2-RC3 using a kernel with INVARIANTS enabled. I was starting pkt-gen on an interface with a large number of rings and had forgotten to increase the if_size sysctl. A memory allocation failed and the system hit a panic.

Panic message:
```
Lock (&nm_mem)->nm_mtx not exclusively locked @ /usr/src/sys/dev/netmap/netmap_mem2.c:271
```


jhk098:
Good catch!

giuseppelettieri:
Thanks. I have removed 'freebsd' from the commit message since it also affects Linux.

PR #736, opened 2020-11-02
Fix BUG_ON() in Linux on vxlan/gretap/hyper-v interfaces

socketpair:
Wrong headroom calculations lead to pskb_expand_head() in the drivers.
It fails since skb is shared. Correct calculations prevent
expanding skb.


socketpair:
@luigirizzo I think this is very important.

giuseppelettieri:
Merged, thanks!

Issue #688, opened 2020-04-11
Cannot allocate memory for second interface


denisSal:
Hi!

Running two instances of any netmap application on two different interfaces on Ubuntu 18.04.4 kernel 4.15.0-96-generic fails (Ryzen 7 3700X CPU on a X470 type motherboard and two Intel X520 NICs) with "Cannot allocate memory". I've found the same issues online and the solution was most often playing with buf/ring/ifc size/num parameters, but nothing I changed helped me and dmesg was not giving any info or errors. After
`echo 1 > /sys/module/netmap/parameters/verbose`
and running two netmap instances again, the same error occured, but dmesg also showed:
`nm_mem_assign_group       iommu group mismatch: 17 vs 19`

Disabling IOMMU solved this for me, by:
Setting `GRUB_CMDLINE_LINUX_DEFAULT="iommu=off"` in `/etc/default/grub`, running `update-grub2` and rebooting.

Just putting this here if someone has the same problem, but is there a better solution?

Thank you,
Denis

Winters123:
don't have the exact same problem but when I launched two netmap-based instances (occupying 100000 buffer numbers in total), the `dmesg` reported: 
```
[5783791.579411] 046.408974 [1023] netmap_obj_malloc         no more netmap_buf objects
[5783791.579415] 046.408979 [1192] netmap_new_bufs           no more buffers after 0 of 512
[5783791.579418] 046.408982 [1944] netmap_mem2_rings_create  Cannot allocate buffers for RX_ring
[5783796.679088] 051.508703 [1174] netmap_extra_free         breaking with head -1
```
I played with `buf_curr_num`, but nothing happened.

giuseppelettieri:
@Winters123: ```buf_curr_num``` is an output parameter that show the currently preallocated netmap buffers. You should use ```buf_num``` instead.

Winters123:
@giuseppelettieri I see. Thanks.

kogdenko:
Hello.
I have same problem
```bash
nm_mem_assign_group       iommu group mismatch: 17 vs 18
```
It is occurred because netmap use same memory allocator (nm_mem in netmap_mem.c) for all physical interfaces running in netmap mode. After assigning iommu group (nm_mem_assign_group)
this allocator can't be used for interfaces from another iommu group.

JDarnley:
[...]

PR #750, opened 2021-01-10
Enable netmap to run on several interfaces from different iommu groups

kogdenko:
mem: create allocator for each iommu group (on demand)
  
  Assign allocator iommu group id (nm_grp) at initialization stage and do not change it later.
  Check group only for hw (hardware) adapters. Virtual adapters can use any
  particular allocator. For example, we can choose vale allocator through vale-ctl -m
  nm_iommu_group_id() return -1 on error instead of 0, because 0 is valid iommu group.

issue: https://github.com/luigirizzo/netmap/issues/688

### DIFF
--- a/LINUX/netmap_linux.c
+++ b/LINUX/netmap_linux.c
@@ -326,11 +326,11 @@ int nm_iommu_group_id(struct device *dev)
 	int id;
 
 	if (!dev)
-		return 0;
+		return -1;
 
 	grp = iommu_group_get(dev);
 	if (!grp)
-		return 0;
+		return -1;
 
 	id = iommu_group_id(grp);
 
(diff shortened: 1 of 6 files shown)


giuseppelettieri:
I like the idea, but why redefining ```netmap_mem_get()```? The only purpose of that function was reference counting.

kogdenko:
netmap_mem_get() hasn't changed. I just added new function netmap_mem_get_iommu() - which create new allocator or add refcount to existing allocator. May be it is better to call it function netmap_mem_iommu_new (like netmap_mem_pt_guest_new) or something. If you have any idea what name is more suitable - let's rename it.

giuseppelettieri:
Ah, OK. The name is fine, I was confused by the fact that in ```netmap_mem_get_iommu()```, line 1790, you call thye cumbersome ```__netmap_mem_get()``` instead of ```netmap_mem_get()```. Apart from that, I think we can merge this.

kogdenko:
Yes and I forget to set appropriate netmap_obj_pool.name for just created allocator. I will fix this.

kogdenko:
Done

PR #762, opened 2021-02-15
mlx5: Update to mlx5 v5.2

kierank:


kierank:
@giuseppelettieri Do you have any thoughts on how we can have a simple test suite for these drivers? Maybe a program that transmits on port 1 and receives on port 2?

kierank:
@Jdarnley Can you update this patch with your change

giuseppelettieri:
Hi @kierank, there already is a testing infrastructure. You can run unit tests with `make unitest` and integration tests with `make intest`. ATM, however, the tests only exercise the "software" ports, that are guaranteed to be available. Maybe we can add a `make `_driver_`-test` (e.g., `make mlx5-test`) target for people who know that they have the hardware available and configured for testing.

kierank:
I think this is ok to merge

PR #771, opened 2021-02-26
mlx5: copy fix for RX packet loss from f14770c7 into 5.1

JDarnley:
Full commit hash f14770c7f84985fccd7a3f28b5ccf167f38ac587

@driesdewinter Do you want to be set as the author of this patch since it is your fix?


driesdewinter:
> @driesdewinter Do you want to be set as the author of this patch since it is your fix?

Thanks for offering but no, that is not needed as far as I'm concerned.

PR #784, opened 2021-03-31
mlx5: always leave one free slot in rx ring

driesdewinter:
Previously mlx5e_netmap_rxsync() could completely fill up the rx ring but
user space cannot tell the difference between empty and completely full, so
it would be considered as empty and the ring would be stuck forever.

This is now avoided by never making more than (ring size - 1) buffer slots
available for packet reception.

### DIFF
--- a/LINUX/mlx5_netmap_linux.h
+++ b/LINUX/mlx5_netmap_linux.h
@@ -368,6 +368,7 @@ int mlx5e_netmap_rxsync(struct netmap_kring *kring, int flags) {
   u_int nm_i = 0; /* index into the netmap ring */
   u_int const lim = kring->nkr_num_slots - 1;
   u_int const head = kring->rhead;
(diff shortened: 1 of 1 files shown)


kierank:
@JDarnley When you have a chance can you test this on our side. 
I don't think we ever reach empty/full though

JDarnley:
I've been using this commit today and I haven't seen any problems or changes in how the RX on the nic behaves.  But as we've said we probably don't reach full because we are ready to handle RX before we get any packets.

giuseppelettieri:
The commit is consistent with what we do with the other drivers, so I am going to merge it if there are no objections. Thanks!

driesdewinter:
Friendly reminder: I see no objections, so it can be merged in then?

giuseppelettieri:
Yep, merged. Thanks!

PR #828, opened 2022-01-09
Paste

micchie:
Good shape in both Linux (tested in kernel 5.11) and FreeBSD (tested in CURRENT snapshot on 06/01/22, need to add the netmap_paste.c entry in sys/conf/files).


micchie:
resubmitting.

PR #830, opened 2022-01-10
Paste submit

micchie:
Good shape in both Linux (tested in kernel 5.11) and FreeBSD (tested in CURRENT snapshot on 06/01/22, need to add the netmap_paste.c entry in sys/conf/files). The first commit (nm_os_ifnet_lock removal) is irrelevant to paste.


giuseppelettieri:
I see that there are some small fixes mixed with the paste code. I'd rather merge those separately (I'll add a comment to them in the "Files changed" section). Moreover, I'd feel safer if the changes to core files were protected with 'WITH_PASTE', so that nothing is changed when paste is not compiled in.

micchie:
Thanks! I will split the commits and resubmit the merge request.

Cheers,
- Michio

Issue #816, opened 2021-09-17
Compilation of i40e fails on (Ubuntu) kernel 5.8.0 and 5.11.0 - MODPOST error


tumtom:
Hi,

I tried compiling netmap with the external i40e versions (2.4.6, 2.12.6, 2.13.10, 2.14.13, 2.15.9, 2.16.11) against all the Ubuntu Kernel versions 5.8.0-XX-generic and 5.11.0-XX-generic.

The make of i40e failed for all versions mentioned above. The error is always at the MODPOST step of i40e:

```bash
if [ -d i40e ] && [ -e Module.symvers ]; then cp Module.symvers i40e; fi
make -C i40e CFLAGS_EXTRA=" -I/home/<username-removed>/netmap-setup/netmap/LINUX -I/home/<username-removed>/netmap-setup/netmap/LINUX -I/home/<username-removed>/netmap-setup/netmap/LINUX/../sys -I/home/<username-removed>/netmap-setup/netmap/LINUX/../sys/dev -DCONFIG_NETMAP -Wno-unused-but-set-variable -g -DCONFIG_NETMAP_NULL -DCONFIG_NETMAP_PTNETMAP -DCONFIG_NETMAP_GENERIC -DCONFIG_NETMAP_MONITOR -DCONFIG_NETMAP_PIPE -DCONFIG_NETMAP_VALE -DNETMAP_DRIVER_SUFFIX=-netmap"  NETMAP_DRIVER_SUFFIX=-netmap KSRC=/lib/modules/5.8.0-63-generic/build
make[1]: Entering directory '/home/<username-removed>/netmap-setup/netmap/LINUX/i40e-2.12.6/src'
*** The target kernel has CONFIG_MODULE_SIG_ALL enabled, but
*** the signing key cannot be found. Module signing has been
*** disabled for this build.
make[2]: Entering directory '/usr/src/linux-headers-5.8.0-63-generic'
  CC [M]  /home/<username-removed>/netmap-setup/netmap/LINUX/i40e-2.12.6/src/i40e_main.o
  CC [M]  /home/<username-removed>/netmap-setup/netmap/LINUX/i40e-2.12.6/src/i40e_ethtool.o
  CC [M]  /home/<username-removed>/netmap-setup/netmap/LINUX/i40e-2.12.6/src/i40e_adminq.o
  CC [M]  /home/<username-removed>/netmap-setup/netmap/LINUX/i40e-2.12.6/src/i40e_common.o
[...]

PR #839, opened 2022-02-10
Fixes ERROR: modpost errors when running make install on Debian Bullseye

badagent:
Similar to closed issue #816 but this time errors occurred when running make install.


giuseppelettieri:
Thanks!

PR #840, opened 2022-02-15
Fix variable 'err' usage and compilation

ocochard:
Fix compilation and correct usage of the err variable:

nmreplay.c:1105:13: error: variable 'err' set but not used [-Werror,-Wunused-but-set-variable]
        int ch, i, err=0;


giuseppelettieri:
Thanks!

PR #841, opened 2022-02-17
Now netmap supports intel ice nic driver

mr-unstoppable:
I have added netmap supports intel ice nic driver

### DIFF
--- a/LINUX/ice_netmap_linux.h
+++ b/LINUX/ice_netmap_linux.h
@@ -1,0 +1,571 @@
+#include <bsd_glue.h>
+#include <net/netmap.h>
+#include <dev/netmap/netmap_kern.h>
+
+int ice_netmap_txsync(struct netmap_kring *kring, int flags);
+int ice_netmap_rxsync(struct netmap_kring *kring, int flags);
+
+extern int ix_crcstrip;
+
+#ifdef NETMAP_LINUX_ICE_PTR_ARRAY
+#define NM_ICE_TX_RING(a, r)		((a)->tx_rings[(r)])
+#define NM_ICE_RX_RING(a, r)		((a)->rx_rings[(r)])
+#else
+#define NM_ICE_TX_RING(a, r)		(&(a)->tx_rings[(r)])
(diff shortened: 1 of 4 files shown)


giuseppelettieri:
Nice, thanks. I don't have the hardware to test this, but a new driver is always welcome. There a few compilation-related things that needs to be fixed, though. I'll add comments below the relevant lines.

LINUX/ice_netmap_linux.h
```diff
@@ -0,0 +1,571 @@
...
+#define NM_ICE_TX_RING(a, r)		(&(a)->tx_rings[(r)])
+#define NM_ICE_RX_RING(a, r)		(&(a)->rx_rings[(r)])
+#endif
+#ifdef NETMAP_LINUX_ICE_PTR_STATE
+#define NM_ICE_STATE(pf)		(&(pf)->state)
+#else
+#define NM_ICE_STATE(pf)		((pf)->state)
+#endif
+
+#ifdef NETMAP_ICE_MAIN
```
giuseppelettieri, line 23: The situation here is similar to i40e, where the `i40e_netmap_linux.h` header is included in two files, each needing some common definitions and a subset of the header's functions. Essentially, `i40e_netmap_linux.h` is two headers in one: the sub-header intended for `i40e_main.c` is selected by the `NETMAP_I40E_MAIN` macro. 

Your header, however, is included in _three_ ice files (`ice_base.c`, `ice_lib.c` and `ice_txrx.c`), so you should split it in three parts. Right now, the compilation of `ice_lib.c` complains about defined but not used functions. Please introduce a `NETMAP_ICE_LIB`, besides `NETMAP_ICE_MAIN`, to skip the attach/configure function definitions when the header is included in `ice_lib.c`. Maybe also rename `NETMAP_ICE_MAIN` to `NETMAP_ICE_BASE`.

giuseppelettieri (changes requested):

giuseppelettieri:
Good, thanks!

PR #853, opened 2022-05-17
Swap the src and dst IPv4 address in the pong msg

steven-chien:
The current ```pkt-gen``` tool only flips the source and destination MAC address without flipping the IP addresses. This becomes problematic when working with links with NAT or hops in between. The PR flips the IPv4 address in a pong response.

### DIFF
--- a/apps/pkt-gen/pkt-gen.c
+++ b/apps/pkt-gen/pkt-gen.c
@@ -1547,6 +1547,11 @@ pong_body(void *data)
 				dpkt[3] = spkt[0];
 				dpkt[4] = spkt[1];
 				dpkt[5] = spkt[2];
+				/* swap source and destination IPv4 */
+				dpkt[13] = spkt[15];
+				dpkt[14] = spkt[16];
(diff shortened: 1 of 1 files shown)


jhk098:
I would at least check that ethertype is 0x8000, and make the swap only in that case.

steven-chien:
Added a check on the ethertype, and also added a warning message that IPv6 is not fully supported in ping-pong since there are no IP options for the timestamp.

apps/pkt-gen/pkt-gen.c
```diff
@@ -1547,7 +1556,15 @@ pong_body(void *data)
 				dpkt[3] = spkt[0];
 				dpkt[4] = spkt[1];
 				dpkt[5] = spkt[2];
+				/* swap source and destination IPv4 */
+				if (ntohs(spkt[6]) & ETHERTYPE_IP) {
```
jhk098, line 1560: s/&/==

jhk098 (approved):

PR #865, opened 2022-07-08
linux/ice: add patch for ice 1.9.7

pkitszel:
Most of the code is originating from [1] by Giuseppe Lettieri.

Source patch [1] was for 1.8.9, last updated in commit e827ea6965d5c47b1:
"(linux/ice: fix hooks and callbacks)"
[1] https://github.com/luigirizzo/netmap/commit/e827ea6965d5c47b13eed24cd7dca2798b40fae9
Style changes applied.

Adjusted for changes in ice 1.9

Signed-off-by: Przemek Kitszel <przemyslaw.kitszel@intel.com>


giuseppelettieri:
Hi, thanks for the patch, but Intel driver patches are auto-generated from a different repository that contains the modified drivers. It would be a nightmare to maintain diffs of patches :) I have now updated the drivers repository and generated patch for 1.9.7.

Your PR also contains an unrelated fix for a typo, which I have committed separately with credits. By the way, for drivers other than ice, that `intel-fix.sh` script tries to fix a syntax error in the Intel `common.mk` makefile. May you take a look at the patch it contains (in `LINUX/intel_fix.sh_`)?

pkitszel:
Thank for your replay.

I have proposed a makefile fix for our ice 1.10 series, but it's unlikely to get into 1.9.

Closing this PR.

PR #869, opened 2022-08-08
Add patch for ice-1.9.11

### LINKED ISSUE #868: add patch for ice-1.9.11
@giuseppelettieri There is next release of ice on sourceforge (1.9.11)

Do you prefer to generate a netmap-enablement patch for it by yourself, or would like to accept contribution?
(I've checked and it should apply cleanly).

Maybe it would also be worthwhile to switch into 'range' mode instead of patch-per-release (as there is such an option implemented already for other drivers/usage).

pkitszel:
Fixes #868


jbrandeb:
fixed the patch to have correct paths.

jbrandeb:
@giuseppelettieri  or - @jhk098 we've been having good luck testing this patch. We found a core netmap issue with 5.14+ kernels not recognizing queue count changes with ethtool -L, but besides that things seem to be working. Can you apply this patch please?

giuseppelettieri:
Hi @jbrandeb and @pkitszel, sorry for the late reply. Patches for the external drivers (at least the ones mantained by us) are generated from another repository, to avoid having to look at diffs of diffs. I have made the `ice` repository public, you can find it [here](https://github.com/netmap-unipi/intel-ice). The `main` branch tracks your releases, while each "`netmap-`_version_" branch contains the patched driver for the "`v`_version_" tag.

I have already created the branch for `v1.9.11` and imported the equivalent of your patch there, in the `netmap-1.9.11` branch. From there I obtained the patch that is pushed on the netmap repository.

pkitszel:
Thanks!

PR #876, opened 2022-09-28
fix ipv6h->ip6_ctlun.ip6_un1.ip6_un1_nxt usage

kimshrier:
This member of the IPv6 header is only 1 byte and therefore should not be byte-swapped using ntohs.  Using ntohs on this value means that none of the cases in the switch statement will match and the default case will always be chosen.

In addition, the correct ICMP definition for IPv6 should be IPPROTO_ICMPV6 instead of IPPROTO_ICMP.


giuseppelettieri:
Merged, thanks!

PR #877, opened 2022-10-03
Add timestamp to netmap_slot

kierank:


giuseppelettieri:
We can't change the size of the slots (ABI break). Maybe we may reuse the `ptr` field when it is not already used for something else (e.g, offsets). Or, we may use offsets to reserve a place for the ts in the netmap buffer. What do you think?

kierank:
Yes, on mlx5 we do exactly that, use `ptr`. But we want to start using ice which has `ptr` being used for offsets.
I'm happy to leave this here for the next ABI change you make because per packet timestamps are useful and now widely available from most modern NICs.

PR #912, opened 2023-03-03
pkt-gen: update IP length, UDP length/checksum when size changes

brian90013:
Hello,

pkt-gen can randomize the packet length using -l maxsize -l minsize arguments. It creates the initial packet using the maximum size, sets the length in the headers, and computes the checksums. Then for each burst it gets a random length and simply truncates the packets to that length.

The result is the IP total length is incorrect as well as the UDP length and checksum (since it is computed over the payload). I have not observed any data sinks that could not handle these inconsistencies, but as a test tool I would like pkt-gen to create valid packets.

I have written a patch that updates the IP/UDP lengths after the new burst size is determined. Then for UDP it recalculates the checksum by walking the new payload. (I considered trying a delta calculation based on the old and new lengths but decided simplicity trumped any performance difference. A second idea would be a checksum lookup table of every possible length computed at startup.) This capability is controlled by a command-line option so it can be used only when desired.

### DIFF
--- a/apps/pkt-gen/pkt-gen.c
+++ b/apps/pkt-gen/pkt-gen.c
@@ -283,6 +283,7 @@ struct glob_arg {
 #define OPT_RANDOM_SRC  512
 #define OPT_RANDOM_DST  1024
 #define OPT_PPS_STATS   2048
+#define OPT_UPDATE_CSUM 4096
 	int dev_type;
 #ifndef NO_PCAP
 	pcap_t *p;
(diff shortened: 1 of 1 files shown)


giuseppelettieri:
Very good, thanks!

apps/pkt-gen/pkt-gen.c
```diff
@@ -1005,6 +1006,85 @@ update_addresses(struct pkt *pkt, struct targ *t)
 	else
 		update_ip6(pkt, t);
 }
+
+static void
+update_ip_size(struct pkt *pkt, struct targ *t, int size)
```
jlduran, line 1011: The compiler whines about `*t` not being unused here and in `update_ip6_size()`.
Should it be removed or mark `__unused`?

PR #914, opened 2023-04-20
Fix compilation on some 32-bit platforms

jlduran:
The type of `tv_sec` is `time_t`, which is 32-bit on some platforms. The type of `tv_nsec` is `long` (`long long` only since C23).

When building the port net/pkt-gen without this patch, a package fallout occurs.

Cast to an `intmax_t` to avoid the warning/error, given `-Werror` (NB: The original patch in the net/pkt-gen port used a `long long`, we opted for an `intmax_t` instead).

See the aftermath of bug report #270440 for further details.

### DIFF
--- a/apps/pkt-gen/pkt-gen.c
+++ b/apps/pkt-gen/pkt-gen.c
@@ -3284,8 +3284,8 @@ out:
 		g.tx_period.tv_nsec = g.tx_period.tv_nsec % 1000000000;
 	}
 	if (g.td_type == TD_TYPE_SENDER)
-	    D("Sending %d packets every  %ld.%09ld s",
-			g.burst, g.tx_period.tv_sec, g.tx_period.tv_nsec);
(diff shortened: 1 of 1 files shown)


jlduran:
From ports Fallout:

armv6, armv7:

    pkt-gen.c:3288:13: error: format specifies type 'long' but the argument has type 'time_t' (aka 'long long') [-Werror,-Wformat]
                            g.burst, g.tx_period.tv_sec, g.tx_period.tv_nsec);
                                     ^~~~~~~~~~~~~~~~~~

i386:

    pkt-gen.c:3288:13: error: format specifies type 'long' but the argument has type 'time_t' (aka 'int') [-Werror,-Wformat]
                            g.burst, g.tx_period.tv_sec, g.tx_period.tv_nsec);
                                     ^~~~~~~~~~~~~~~~~~

giuseppelettieri:
good, thanks!

PR #915, opened 2023-04-21
apps: Import man page changes, mostly from FreeBSD

jlduran:
For bridge.8 and nmreplay.8, sync with FreeBSD.

For pkt-gen.8:
Chiefly, change .Ar (argument) macros with .Cm (command modifier).
Per mdoc(7):

> The arguments to the `Ar` macro are names and placeholders for command arguments; for fixed strings to be passed verbatim as arguments, use `Fl` or `Cm`.

Also, split a long command line.

For lb.8, leave as is, given FreeBSD renamed `vale-ctl` to `valectl`.

### DIFF
--- a/apps/bridge/bridge.8
+++ b/apps/bridge/bridge.8
@@ -23,7 +23,7 @@
 .\"
 .\" $FreeBSD$
 .\"
-.Dd October 23, 2018
+.Dd November 21, 2020
 .Dt BRIDGE 8
 .Os
 .Sh NAME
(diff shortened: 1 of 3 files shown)


giuseppelettieri:
I understand that the references to the papers where not in the correct section, but why remove them completely?

jlduran:
> I understand that the references to the papers where not in the correct section, but why remove them completely?

Good question, I wondered the same when I removed them. It was taken from this commit:
https://github.com/freebsd/freebsd-src/commit/c2231fb0f803df9dc76e1aaacbbdcc32fa1c1c36

I believe because those papers are already referenced in netmap(4).

jlduran:
You are welcome. I'll try down-streaming the changes.
Thank you for the fast response.

PR #925, opened 2023-06-10
add offset support to host rings

msaare123:
Slot offsets were not taken into account in host rings and caused malformed packets were send/received to/from host network stack when offsets were configured to NIC ring slots.


giuseppelettieri:
Looks good, thanks! May you add it to the txsync as well?

msaare123:
> Looks good, thanks! May you add it to the txsync as well?

Which one? netmap_grab_packets() is called from netmap_txsync_to_host()

PR #926, opened 2023-06-15
Make VLAN tags visible to Linux generic adapter

msaare123:
Push tag in skb metadata back to payload so it will be visible in rx ring

### DIFF
--- a/LINUX/netmap_linux.c
+++ b/LINUX/netmap_linux.c
@@ -550,11 +550,21 @@ linux_generic_rx_handler_common(struct mbuf *m)
 	   can see it. */
 	skb_push(m, ETH_HLEN);
 
+	/* First VLAN tag has been already popped to skb metadata. */
+	if (skb_vlan_tag_present(m)) {
+		m = __vlan_hwaccel_push_inside(m);
+	}
+
 	/* Possibly steal the mbuf and notify the pollers for a new RX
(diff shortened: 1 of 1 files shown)


vmaffione:
Looks like a good idea, but we would need to check that it compiles for all the earlier versions on Linux. @giuseppelettieri ?

msaare123:
> Looks like a good idea, but we would need to check that it compiles for all the earlier versions on Linux. @giuseppelettieri ?

What is range of supported kernels? With quick glance this should be compatible from v4.x kernels.

vmaffione:
We actually support Linux since 2.6.34.

The configure script `LINUX/configure` has many tests to check for internal kernel functions and signatures. The script will produce a `netmap_linux_config.h` with all the #define that allow the code to support many versions.
Then there is a `LINUX/scripts/np` that automates the build across all the possible versions, but it's not so easy to setup.

giuseppelettieri:
> We actually support Linux since 2.6.34.

yes, we do, but we should really stop doing that. Anyway, I'll check compilation and report back.

giuseppelettieri:
OK, we can merge this. Then I will push some commits that will fix compilation for all kernels. For very old kernels, however, the vlan feature will simply be disabled, since emulating the missing pieces becomes too cumbersome.

PR #938, opened 2023-10-26
linux: Use correct define for vlan untagging.

Cordell-O:
Define being generated in LINUX/configure was slightly different to what was being used in code.


giuseppelettieri:
Thanks!

PR #940, opened 2023-11-08
libnetmap: remove interface name validation

fichtner:
When trying to use a VLAN device (e.g. "em0.123") with a dot
the library fails to parse the interface correctly. The former
pattern is much too restrictive given that almost all characters
can be coerced into a device name via ifconfig.

Remove the particularly restrictive validation.  Some characters
still cannot be used as an interface name as they are used as
delimiters in the syntax, but this allows to be able to use most
of them without an issue.

Note that the dotted VLAN naming is the default in FreeBSD now
and a patch was proposed here https://reviews.freebsd.org/D42485
as well.

### DIFF
--- a/libnetmap/nmreq.c
+++ b/libnetmap/nmreq.c
@@ -156,11 +156,6 @@ nmreq_header_decode(const char **pifname, struct nmreq_header *h, struct nmctx *
 	for (pipesep = vpname; pipesep != scan && !index("{}", *pipesep); pipesep++)
 		;
 
-	if (!nm_is_identifier(vpname, pipesep)) {
-		nmctx_ferror(ctx, "%s: invalid port name '%.*s'", *pifname,
-				pipesep - vpname, vpname);
-		goto fail;
-	}
 	if (pipesep != scan) {
 		pipesep++;
 		if (*pipesep == '\0') {


fichtner:
Any thoughts on this? For default VLANs this is basically broken in FreeBSD 13.2 and 14.0 at the moment.

giuseppelettieri:
Hi @fichtner, sorry but I have not had time to look at this until now. `nm_is_identifier()` is used in three places for three different purposes: 1) check the part after a `vale` prefix; 2) check the interface name; 3) check the pipe name after a curly bracket. I don't think we really need to change 1 and 3, so we should either create another `nm_is_valid_interface()` just for 2, or maybe just remove the check in 2.  I think the latter is preferable, if there is really nothing to check in interface names. What's your opinion?

fichtner:
@giuseppelettieri thanks for responding. You mean this one?

https://github.com/luigirizzo/netmap/blob/6055e3b0e20b7d93d92cc983fd08155d4078f3ac/libnetmap/nmreq.c#L159-L163

I haven't tested but it looks reasonable to remove it in order to fix this. The special character situation still applies, but I reckon it will complain later anyway when the "port" is not found because it had a special character like "}". But I also think that in that case people should reevaluate the device name policy. :)

giuseppelettieri:
yes, that one. If you want to create a new pull request I will merge it.

fichtner:
Ok, let me test this today and update the PR then.

fichtner:
@giuseppelettieri already confirmed... updated the PR accordingly

PR #946, opened 2024-01-07
linux: use net_if_rx_ni() in nm_os_send_up()

msaare123:
In older kernels net_if_rx() is to be used in interrupt context and netmap nm_os_send_up() is from process context. Thus using netif_rx_ni() is correct function variant in this case and using it prevents pending softirqs. In newer kernels netif_rx() can handle both contexts.

### DIFF
--- a/LINUX/configure
+++ b/LINUX/configure
@@ -1305,6 +1305,16 @@ EOF
 	}
 EOF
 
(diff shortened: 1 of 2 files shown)


giuseppelettieri:
Nice, thanks!

Issue #937, opened 2023-08-24
kernel updated, pkt-gen cause kernel panic


ke-ding:
archlinux, kernel version updated to 6.4.11-arch2-1. rebuild netmap, pkt-gen cause kernel panic

[  344.065162] detected buffer overflow in strcpy
[  344.065172] ------------[ cut here ]------------
[  344.065173] kernel BUG at lib/string_helpers.c:1027!
[  344.065177] invalid opcode: 0000 [#1] PREEMPT SMP NOPTI
[  344.065180] CPU: 4 PID: 1451 Comm: pkt-gen Tainted: G           OE      6.4.11-arch2-1 #1 97b2f722c7732577cb713428e1f14bfdbe1faa91
[  344.065183] Hardware name: Intel(R) Client Systems NUC10i7FNK/NUC10i7FNB, BIOS FNCML357.0038.2020.0131.1422 01/31/2020
[  344.065184] RIP: 0010:fortify_panic+0x13/0x20
[  344.065189] Code: 41 5d c3 cc cc cc cc 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 f3 0f 1e fa 48 89 fe 48 c7 c7 b8 f7 26 ba e8 2d 50 b2 ff <0f> 0b 66 66 2e 0f 1f 84 00 00 00 00 00 90 90 90 90 90 90 90 90 90
[  344.065191] RSP: 0018:ffffade0029ff958 EFLAGS: 00010246
[  344.065194] RAX: 0000000000000022 RBX: 0000000000000000 RCX: 0000000000000027
[  344.065195] RDX: 0000000000000000 RSI: 0000000000000001 RDI: ffff8b2d20d216c0
[  344.065196] RBP: ffffade0029ffaf0 R08: 0000000000000000 R09: ffffade0029ff800
[  344.065197] R10: 0000000000000003 R11: ffffffffbaaca868 R12: ffffade0029ff99c
[  344.065199] R13: ffff8b2584030000 R14: 00000000ffffffff R15: 0000000000000000
[  344.065200] FS:  00007f16286dc740(0000) GS:ffff8b2d20d00000(0000) knlGS:0000000000000000
[  344.065202] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[  344.065203] CR2: 00007f1628531c70 CR3: 0000000141ad8003 CR4: 00000000003706e0
[  344.065204] Call Trace:
[  344.065206]  <TASK>
[...]

PR #948, opened 2024-02-27
Fix Linux 6.5 support to avoid fortify panic

brian90013:
Hello,

I believe this patch is a fix for issue #937 where fortify causes a panic in `tc_configure()` due to a "detected buffer overflow in strcpy". The `nlattr *` is set to `&nlreq.hdr` which has a fixed size causing the overflow. By setting `nlattr *` to `&nlreq` instead, the code sees the 100 byte `buf` space and does not panic. I tested this on kernel 6.5.0-21 and verified the `qdisc` name was properly set by using `tc`. With the fix I could run `pkt-gen` on multiple emulated netmap interfaces without seeing a panic.

While here I removed a duplicate `strcpy()` in `netmap_sink_init()`. I assumed "nmsink" is preferred to "nmsink%d".

### DIFF
--- a/LINUX/netmap_linux.c
+++ b/LINUX/netmap_linux.c
@@ -850,7 +850,7 @@ tc_configure(struct ifnet *ifp, const char *qdisc_name,
 	}
 
 	/* Push TCA_KIND attr. */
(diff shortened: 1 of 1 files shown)


giuseppelettieri:
thanks!

PR #954, opened 2024-04-03
Some mellanox housekeeping and a new driver version

JDarnley:
* An old bugfix was missed in driver version 5.3.  I've had this patch for 2 years and I think I tested it at the time but we have moved past that version.
* I discovered that mellanox/nvidia provide a smaller tarball we can download to build netmap that does not bundle large debs.  This smaller tarball has the same sha256sum as the first inner tarball of the larger one.
* We are starting to test driver version 5.8 but early tests show no obvious problems with TX or RX.  If others want to test it please do.
* We use dkms to package but never shared the "recipe".


giuseppelettieri:
Looks good, thanks!

Issue #972, opened 2025-02-11
build failure due to upstream linux change


cpackham:
Hi,

We just bumped our linux kernel from 6.6.75 to 6.6.76 and netmap started failing to build.

```
awplus-arx200,linux: In file included from /usr/src/netmap-974b68c0/LINUX/../sys/dev/netmap/netmap_vale.c:58:
awplus-arx200,linux: /usr/src/netmap-974b68c0/LINUX/../sys/dev/netmap/netmap_vale.c: In function 'netmap_vi_create':
awplus-arx200,linux: /usr/src/netmap-974b68c0/LINUX/bsd_glue.h:345:26: error: assignment to 'struct ax25_dev *' from incompatible pointer type 'struct netmap_adapter *' [-Werror=incompatible-pointer-types]
awplus-arx200,linux:   345 |         (_ifp)->ax25_ptr = _na;                                         \
awplus-arx200,linux:       |                          ^
awplus-arx200,linux: /usr/src/netmap-974b68c/LINUX/../sys/dev/netmap/netmap_kern.h:1723:9: note: in expansion of macro 'if_setnetmapadapter'
awplus-arx200,linux:  1723 |         if_setnetmapadapter(ifp, na);                                   \
awplus-arx200,linux:       |         ^~~~~~~~~~~~~~~~~~~
awplus-arx200,linux: /usr/src/netmap-974b68c0/LINUX/../sys/dev/netmap/netmap_vale.c:1476:9: note: in expansion of macro 'NM_ATTACH_NA'
awplus-arx200,linux:  1476 |         NM_ATTACH_NA(ifp, &vpna->up);
awplus-arx200,linux:       |         ^~~~~~~~~~~~
```

Looks like this is due to the upstream change https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-6.6.y&id=7705d8a7f2c26c80973c81093db07c6022b2b30e which changed `net_device::ax25_ptr` from a `void *` to `struct ax25_dev __rcu *`

PR #975, opened 2025-03-26
Cast the network_adapter pointer when saving in ax25_ptr

brian90013:
This was reported in issue #972 and also seen by me upgrading from kernel 6.1.0-31-amd64 to 6.1.0-32-amd64 on Debian 12. The `netmap.ko` driver fails to compile because the `struct net_device` member `ax25_ptr` was modified in Linux kernel commit [7705d8a7f2c2](https://web.git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-6.6.y&id=7705d8a7f2c26c80973c81093db07c6022b2b30e) and is now type `struct ax25_dev *` not `void *` causing the assignment of `struct network_adapter *` to fail.

This MR casts `struct network_adapter *` to `(void *)` before assigning to `ax25_ptr`. My build and test with `pkt-gen` both work after this modification. Many thanks to the issue creator for finding the exact upstream commit.

### DIFF
--- a/LINUX/bsd_glue.h
+++ b/LINUX/bsd_glue.h
@@ -300,7 +300,7 @@ struct thread;
  * for netmap-capable is some magic in the area pointed by that.
  */
 #define if_setnetmapadapter(_ifp, _na)	do { 				\
-	(_ifp)->ax25_ptr = _na;						\
+	(_ifp)->ax25_ptr = (void *)_na;					\
 } while (0)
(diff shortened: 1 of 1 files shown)


brian90013:
Came back to this in a new week and was unable to cause the build failure (without the patch). Tried to track down the change but unable to at the moment.

